PR #20, opened 2021-11-01
PyInPort Implementation

mgkwill:


lava/magma/core/model/py/ports.py
```diff
@@ -51,27 +57,74 @@ def peek(self) -> np.ndarray:
...
+        # Draf Impl. Receives Vector from Sparse
+        #
+        # if not from PyOutPortVectorSparse we need to
+        # process the data received
+        if self._csp_port:
+            return (self._csp_port.recv(),
+                    self._csp_port.recv())
+        else:
+            return (np.zeros(self._shape, self._d_type),
+                    np.zeros(self._shape, self._d_type))
```
mgkwill, line 75: In the case that PyInPortVectorSparse receives data from anything other than PyOutPortVectorSparse we need to understand what is received and how to process it. Should the compiler expose that information?

@awintel

tests/magma/core/model/py/test_ports.py
```diff
@@ -0,0 +1,65 @@
...
+
+            send_py_port.send(data)
+
+            assert np.array_equal(data, recv_py_port.recv())
+        finally:
+            smm.shutdown()
+
+    # TODO: Need to test the process_builder.build flow which creates the
+    # TODO: inport and outports within the function.
+
```
mgkwill, line 62: Add test of process_builder.build flow to one of the process builder tests.

tests/magma/core/model/py/test_ports.py
```diff
@@ -0,0 +1,65 @@
...
+def get_channel(smm, data, size, name="test_channel") -> PyPyChannel:
+    channel = PyPyChannel(
+        smm=smm, src_name=name, dst_name=name, shape=data.shape,
+        dtype=data.dtype, size=size
+    )
+    return channel
+
+
+class TestPyPorts(unittest.TestCase):
+    def test_port_send_out_vec_dense_to_in_vec_dense(self):
```
mgkwill, line 25: Change name of test method?

Add tests for 
- [ ] PyInPortVectorSparse
- [ ] PyInPortScalarDense
- [ ] PyInPortScalarSparse

tests/magma/core/model/py/__init__.py
```diff
@@ -0,0 +1 @@
+__import__("pkg_resources").declare_namespace(__name__)
```
mgkwill, line 1: Do we need a license here?

lava/magma/core/model/py/ports.py
```diff
@@ -51,27 +57,74 @@ def peek(self) -> np.ndarray:
 
 
 class PyInPortVectorSparse(PyInPort):
+    """Python implementation of Vector Sparse InPort
+    """
+    # sparse port recieves twice, for val and index.
+    # if not sparse port we need to convert to dense,
+    # vice-versa, possible scipy or numpy tools for conversion
```
mgkwill, line 64: Remove on final version, convert to docstring

lava/magma/core/model/py/ports.py
```diff
@@ -51,27 +57,74 @@ def peek(self) -> np.ndarray:
...
     def recv(self) -> ty.Tuple[np.ndarray, np.ndarray]:
-        pass
+        # Draf Impl. Receives Vector from Sparse
+        #
+        # if not from PyOutPortVectorSparse we need to
+        # process the data received
```
mgkwill, line 69: Remove all of these comments and convert to docstring as appropriate.

mgkwill:
pypychannel implimentation of CspSendPort does not appear to support sending data, idx as of now:
https://github.com/lava-nc/lava/blob/4adca2feea5c4ac28f9522c0fcc97a766ff8813e/lava/magma/compiler/channels/pypychannel.py#L99-L108

However that appears to be expected in mockup:
https://github.com/lava-nc/lava/blob/4adca2feea5c4ac28f9522c0fcc97a766ff8813e/lava/magma/core/model/py/ports.py#L107-L117

lava/magma/core/model/py/ports.py
```diff
@@ -51,27 +57,74 @@ def peek(self) -> np.ndarray:
...
+            return (self._csp_port.recv(),
+                    self._csp_port.recv())
+        else:
+            return (np.zeros(self._shape, self._d_type),
+                    np.zeros(self._shape, self._d_type))
 
     def peek(self) -> ty.Tuple[np.ndarray, np.ndarray]:
-        pass
+        if self._csp_port:
+            return (self._csp_port.peek(),
```
joyeshmishra, line 79: This is incorrect. It might not be possible to do a peek twice. Let's talk to Harry and Andreas on how to solve this.

lava/magma/core/model/py/ports.py
awintel: We need to actually implement the messaging protocol described in the powerpoint slides. Each send/recv function should be compatible with any other recv/send function.

The message header should identify from what type of class it was send from. Perhaps, given the knowledge what Port connects to what other port, it could provide further information to the port classes for them to interpret the received messages.

lava/magma/core/model/py/ports.py
awintel: Peeks don't remove the data. Hence, you can call it as often as you want.

lava/magma/core/model/py/ports.py
```diff
@@ -55,16 +59,25 @@ def peek(self) -> np.ndarray:
 class PyInPortVectorSparse(PyInPort):
     """Python implementation of Vector Sparse InPort
     """
+    # sparse port recieves twice, for val and index.
```
awintel, line 62: I don't think it needs to receive twice. The sender should probably interleave index and value vectors as described in pptx. This way both vector-valued and scalar-valued recv() functions can deal with the data.

Probably best to write unit tests first to convince yourself of the usage model.

lava/magma/core/model/py/ports.py
```diff
@@ -55,16 +59,25 @@ def peek(self) -> np.ndarray:
...
+    # if not sparse port we need to convert to dense,
+    # vice-versa, possible scipy or numpy tools for conversion
     def recv(self) -> ty.Tuple[np.ndarray, np.ndarray]:
+        # Draf Impl. Receives Vector from Sparse
+        #
+        # if not from PyOutPortVectorSparse we need to
+        # process the data received
         if self._csp_port:
-            return self._csp_port.recv()
+            return (self._csp_port.recv(),
```
awintel, line 71: indies and values must be interleaved.

awintel (changes requested):
Maybe best to start with unit tests for all combinations of {VecDense, VecSparse, ScaDense, ScaSparse). 
We can also have unit tests for the packed message formats. This may help clarify the messaging protocol.

mgkwill:
source now moved from /ava to /src/lava

mgkwill:
Looks like last commit introduced some merge errors. Will push a fix.

awintel (commented):
Just some observations since you have not asked for a review yet:
- A lot of the test code for the test_send_something_to_something() functions are pretty repetitive. You should factor most of the repeated code out into a function which which you could call with different class arguments and such...
- You might even be able to condense everything down into a single test that merely iterates over a list of Port-Port-type combinations which you then pass into a helper function that performs the actual channel transaction. 
- It might be wise to seed your random number generator so you get repeatable results each time.
- Right now there are print statements in the code. Be sure to remove them in the end. Unit tests should always be silent and not print anything after debugging is over. You can then further looper over this helper function and repeat the same test sequence with Ports of different shape rather than explicitly typing all this out.

mgkwill:
> Just some observations since you have not asked for a review yet:
> 

Hi @awintel thanks for the feedback. I'm open to review/feedback at every stage of the process, so feel free to chime in at any time.

> * A lot of the test code for the test_send_something_to_something() functions are pretty repetitive. You should factor most of the repeated code out into a function which which you could call with different class arguments and such...
> * You might even be able to condense everything down into a single test that merely iterates over a list of Port-Port-type combinations which you then pass into a helper function that performs the actual channel transaction.

I planned to refactor these unit tests. I'm focused right now on understanding the transformations from sparse/dense/vector/scalar and getting working code done. I don't mind putting my process and progress out there and making sure I check in often even in a state that is not final. I'll go back over unit tests and code and refactor before final code check in and review. 

> * It might be wise to seed your random number generator so you get repeatable results each time.

Agreed.

> * Right now there are print statements in the code. Be sure to remove them in the end. Unit tests should always be silent and not print anything after debugging is over. You can then further looper over this helper function and repeat the same test sequence with Ports of different shape rather than explicitly typing all this out.

Agreed.

Thanks again for the feedback.

mgkwill:
Closing in favor of a rework.

PR #24, opened 2021-11-03
Enable RefPort connections

PhilippPlank:
-added RefPorts and VarPorts to find_processes
-added var_ports to compile_proc_models
-channels are now also build for RefPort -> VarPort connections
-added unit tests

Next step is to add RefPorts in the synchronization protocol to allow sending data.

### DIFF
--- a/tests/magma/compiler/test_compiler.py
+++ b/tests/magma/compiler/test_compiler.py
@@ -13,8 +13,9 @@ from lava.magma.core.model.sub.model import AbstractSubProcessModel
 from lava.magma.core.sync.domain import SyncDomain
 from lava.magma.core.sync.protocol import AbstractSyncProtocol
 from lava.magma.core.sync.protocols.async_protocol import AsyncProtocol
-from lava.magma.core.process.ports.ports import InPort, OutPort
-from lava.magma.core.model.py.ports import PyInPort, PyOutPort
+from lava.magma.core.process.ports.ports import (
+    InPort, OutPort, RefPort, VarPort)
+from lava.magma.core.model.py.ports import PyInPort, PyOutPort, PyRefPort
 from lava.magma.core.run_configs import RunConfig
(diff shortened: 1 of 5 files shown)


tests/magma/compiler/test_compiler.py
```diff
@@ -262,6 +267,32 @@ def test_find_process_circular(self):
...
+        p1, p2, p3 = (
+            ProcA(),
+            ProcB(),
+            ProcC()
+        )
+        # Create complicated circular structure with joins and forks
+        p1.ref.connect(p2.var_port)
+        p2.out.connect(p3.inp)
+
+        # Regardless where we start searching...
```
awintel, line 284: Did you just copy paste a unit test and is this still tentative or what is this test really about?

tests/magma/compiler/test_compiler.py
```diff
@@ -658,6 +689,36 @@ def test_create_channel_builders_hierarchical_process(self):
...
         self.assertEqual(chb[0].dst_process, p.procs.proc2)
 
+    def test_create_channel_builders_ref_ports(self):
+        """Checks creation of channel builders when a process is connected
+        using a RefPort to another process."""
+
+        # create a process with a RefPort (source)
+        src = ProcA()
+
+        # create a process with a var (destination)
```
awintel, line 699: Minor thing: But please check for spelling also in comments to be consistent. The first word of a sentence normally starts with a capital letter. See other comments.

awintel (commented):
Just came across this... so leaving some comments.

tests/magma/compiler/test_compiler.py
PhilippPlank: Yes, the test was copied to keep a similar structure and I forgot to change a comment. It tests finding the connected processes if they are connected via a Ref/Var-port.

mgkwill:
Source now moved from /lava to /src/lava.

PR #27, opened 2021-11-04
Issue11: forking/joining ports strawman 

jlakness-intel:
simple implementation for review to promote design discussion

### DIFF
--- a/lava/magma/core/model/py/ports.py
+++ b/lava/magma/core/model/py/ports.py
@@ -5,7 +5,6 @@ import typing as ty
 from abc import abstractmethod
 from enum import Enum
 import functools as ft
-import operator as op
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


lava/magma/core/model/py/ports.py
```diff
@@ -38,16 +40,10 @@ def probe(self) -> bool:
 
 class PyInPortVectorDense(PyInPort):
     def recv(self) -> np.ndarray:
-        if self._csp_port:
-            return self._csp_port.recv()
-        else:
-            return np.zeros(self._shape, self._d_type)
+        return ft.reduce(lambda acc,csp_port:acc+csp_port.recv(),self._csp_ports,np.zeros(self._shape, self._d_type))
```
awintel, line 43: This might exceed the 80 character line limit but good to reformat this for readability anyways.

awintel (approved):
Thanks for getting this started. Would be good to have a dedicated unit test that assigns multiple CspPorts to a PyPort.

mgkwill (approved):
There are a number of linting issues, but once fixed, this looks good to me.
https://github.com/lava-nc/lava/runs/4107261786?check_suite_focus=true

mgkwill:
I think it would be good to have a simple implementation and build complexity in further changes.

lava/magma/core/model/py/ports.py
jlakness-intel: I was trying to figure out how to test the builder similar to this test: 
https://github.com/jlakness-intel/lava/blob/1bd57219993f516204ac21260ef50b8d49594d1b/tests/magma/core/model/test_py_model.py#L273

None of these build from PyPorts, however. They all skip that step and build the ProcessModel.csp_ports dictionary directly rather than go through the compiler to collect them from PyPort._csp_port (renamed _csp_ports). So I'm not actually sure where this would be tested, and what functional outcome should be tested for. Of course I could test for exactly what I know the code is doing, but I still wouldn't know much about whether it works in context.

I also looked for an end-to-end test of multiple Process/Port through the compilation chain that could be adapted to fork/join connections, and couldn't find much there. We should be able to check to see that a fork/join port in a process compiles to a fork/join port in a model, and then actually sends messages through the runtime as expected.

Observation: there is a lot of hidden complexity to the builder, and it's not simple to find where the magic happens because of the use of type variables in the compiler.

joyeshmishra (approved):

PR #31, opened 2021-11-05
issue13: Process Models in C for python and embedded platforms

jlakness-intel:
opening incomplete for ongoing in-code discussion of what this should look like


lava/magma/core/model/c/ports_embedded.c
```diff
@@ -0,0 +1,16 @@
+// implementation of ports API on the embedded platform
+#define EMBEDDED
+#include "ports.h"
+
+int port_send(Port *port,void* data){
```
awintel, line 5: Why not just send, recv, peek, probe for short like in Python API?
There are likely no other functions with the namespace when the user writes this code and the typed Port argument makes it pretty clear what it does.

lava/magma/core/model/c/ports.h
```diff
@@ -0,0 +1,26 @@
+#ifndef _PORTS_H
+#define _PORTS_H
+#include <stdint.h>
+
+#ifdef PYTHON
+#include <Python.h>
+typedef PyObject Port;
+#else // HOST or EMBEDDED
```
awintel, line 8: HOST and EMBEDDED likely require different treatment. 

On EMBEDDED you will purely use Andrew's CSP channel implementation while on HOST you might use Andrew's CSP API and may have to communicate with the SH either somehow via GRPC or via Ray in the future.

awintel (commented):
Just providing some general inline feedback...

awintel (commented):
Just providing some general inline feedback...

awintel (commented):
Function and module documentation would be useful to emphasize design intent.

awintel (commented):
Can you share a more end-to-end mock-up example or suite of TDD unit tests to show case the API?

lava/magma/core/model/c/ports_embedded.c
jlakness-intel: changed in https://github.com/lava-nc/lava/pull/31/commits/1caad0cbc61ce87d61ddb10a03b4ae18b506edc6

lava/magma/core/model/c/ports.h
jlakness-intel: Yes. Haven't gotten there yet.

lava/magma/core/model/c/ports_python.c
```diff
@@ -0,0 +1,43 @@
+#define PYTHON
+#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
+#include <Python.h>
+#include "numpy/arrayobject.h"
+#include <assert.h>
+#include <stdint.h>
+#include "ports.h"
+
+size_t send(PyObject *port, void *data, size_t m){ // send via pointer
```
jlakness-intel, line 9: might want to change this at some point to get 'm' from the port object. doing it this way for now to make one less thing that could go wrong.

mgkwill:
source now moved from /ava to /src/lava, please update patch accordingly.

lava/magma/core/model/c/model.py
```diff
@@ -1,11 +1,49 @@
...
+
+            setup(
+                configuration=configuration,
+                script_args=["build_ext", "--inplace"],
+            )
+            invalidate_caches()
+            module = import_module("custom")
+            # from custom import Custom
+            bases += (module.Custom, AbstractPyProcessModel)
+        return super().__new__(cls, name, bases, attrs)
```
jlakness-intel, line 43:
```
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

This requires some judo I think, but it should be resolvable. Another possibility is to just compile the object in a separate function and then do manual type creation with `type(...)` rather than the metaclass.

PhilippPlank:
I will close this PR, since a new one was created, please see #152

PR #67, opened 2021-11-17
Disable non-blocking execution

PhilippPlank:
Currently non-blocking execution is not fully implemented. Calling run(..) with blocking=False now raises NotImplementedError with meaningful error message.


awintel (approved):
Thanks for taking care of this.

mathisrichter (approved):
Looks good to me - thanks, Philipp

PR #73, opened 2021-11-18
Conv Process Implementation

bamsumit:
Implementation of conv process and process models.


tests/lava/proc/conv/test_models.py
```diff
@@ -62,6 +62,7 @@ def run_spk(self):
             data = self.data[..., (self.current_ts - 1) % buffer]
             print(f'Sending data ={data[data!=0]}')
         self.s_out.send(self.data[..., (self.current_ts - 1) % buffer])
+        self.s_out.flush()
```
awintel, line 65: This command does not do anything right now. And once it does, you will have to modify the Process a bit to not send dummy data in the end.

src/lava/proc/conv/process.py
```diff
@@ -0,0 +1,104 @@
...
+        convolution dilation. Default is 1.
+    groups : int
+        number of groups in convolution. Default is 1.
+
+    Note
+    ----
+    padding, stride and dilation are expected in (X, Y) or (W, H) if tuple.
+    """
+    def __init__(self, **kwargs):
+        # the process In/OutPort shapes are considered to be XYZ(WHC) format.
```
awintel, line 38: Sentences usually start with a capital letter.

src/lava/proc/conv/process.py
```diff
@@ -0,0 +1,104 @@
...
+
+from lava.magma.core.process.process import AbstractProcess
+from lava.magma.core.process.variable import Var
+from lava.magma.core.process.ports.ports import InPort, OutPort
+
+from lava.proc.conv import utils
+
+
+class Conv(AbstractProcess):
+    """Convolution connections between neurons.
```
awintel, line 15: Convolutional connections between neurons.
or 
Convolution connection process between neurons.

src/lava/proc/conv/process.py
```diff
@@ -0,0 +1,104 @@
...
+    padding, stride and dilation are expected in (X, Y) or (W, H) if tuple.
+    """
+    def __init__(self, **kwargs):
+        # the process In/OutPort shapes are considered to be XYZ(WHC) format.
+        # the kernel weight shape is expected to be in (C_out, W, H, C_in)
+        # format.
+        # Why? This is the format that Loihi conv feature uses.
+        super().__init__(**kwargs)
+
+        def broadcast_arg(name, default):
```
awintel, line 44: Broadcasting should be done by the Compiler/ProcModelBuilder when it expands the Var(..., init=...) argument to the Var value at the beginning of execution on the remote process executing a particular ProcModel.
This avoids allocating unnecessary memory at the process level.
Is this broadcasting feature insufficient?

src/lava/proc/conv/utils.py
```diff
@@ -0,0 +1,236 @@
...
+                # torch cannot handle negative stride
+                weight[:, ::-1, ::-1].transpose([0, 3, 2, 1]).copy()
+            ),
+            stride=stride[::-1].tolist(),
+            padding=padding[::-1].tolist(),
+            dilation=dilation[::-1].tolist(),
+            groups=groups
+        )[0].cpu().data.numpy().transpose([2, 1, 0])
+    else:
+        output = conv_scipy(
```
awintel, line 160: Would numpy.convolve also work? That way we might be able to reduce mandatory requirements for /lava

src/lava/proc/conv/models.py
```diff
@@ -0,0 +1,81 @@
...
+        s_in = self.s_in.recv()
+        a_out = utils.conv(
+            s_in, self.weight,
+            self.kernel_size, self.stride, self.padding, self.dilation,
+            self.groups[0]
+        )
+        self.a_out.send(self.clamp_precision(a_out))
+        self.a_out.flush()
+
+    def run_lrn(self):
```
awintel, line 36: You don't have to list methods that are not used.

src/lava/proc/conv/models.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def run_spk(self):
+        s_in = self.s_in.recv()
+        a_out = utils.conv(
+            s_in, self.weight,
+            self.kernel_size, self.stride, self.padding, self.dilation,
+            self.groups[0]
+        )
+        self.a_out.send(self.clamp_precision(a_out))
+        self.a_out.flush()
```
awintel, line 34: Unnecessary. Not implemented right now.

src/lava/proc/conv/models.py
```diff
@@ -0,0 +1,81 @@
...
+class AbstractPyConvModel(PyLoihiProcessModel):
+    """Abstract template implemetation of PyConvModel."""
+    kernel_size: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=8)
+    stride: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=8)
+    padding: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=8)
+    dilation: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=8)
+    groups: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=8)
+
+    def run_spk(self):
+        s_in = self.s_in.recv()
```
awintel, line 27: For clarity one should probably define an abstract property for s_in. Since s_in does not exist yet at the object level. This only happens to not throw a linter warning because there exists a class variable with the same name.
But in reality you are only providing an implementation for s_in and a_out in the sub classes.

src/lava/proc/conv/models.py
```diff
@@ -0,0 +1,81 @@
...
+    """Binary spike fixed point synapse implementation."""
+    s_in: PyInPort = LavaPyType(PyInPort.VEC_DENSE, bool, precision=1)
+    a_out: PyOutPort = LavaPyType(PyOutPort.VEC_DENSE, np.int32, precision=24)
+    weight: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=8)
+
+    def clamp_precision(self, x):
+        return utils.signed_clamp(x, bits=24)
+
+
+# @implements(proc=Conv, protocol=LoihiProtocol)
```
awintel, line 66: Delete dead code.

src/lava/proc/conv/models.py
```diff
@@ -0,0 +1,81 @@
...
+    """Binary spike float synapse implementation."""
+    s_in: PyInPort = LavaPyType(PyInPort.VEC_DENSE, bool, precision=1)
+    a_out: PyOutPort = LavaPyType(PyOutPort.VEC_DENSE, float)
+    weight: np.ndarray = LavaPyType(np.ndarray, float)
+
+
+@implements(proc=Conv, protocol=LoihiProtocol)
+@requires(CPU)
+@tag('fixed_pt')
+class PyConvModelBinaryFixed(AbstractPyConvModel):
```
awintel, line 56: We also need to think about how we support other features to make it "more bit accurate" like wgtExp, variable bit precision, fanout type, etc. These are functionally visible to the user.

tests/lava/proc/conv/test_models.py
```diff
@@ -0,0 +1,304 @@
...
+        self.select_tag = select_tag
+
+    def select(self, proc, proc_models):
+        for pm in proc_models:
+            if self.select_tag in pm.tags:
+                return pm
+        raise AssertionError('No legal ProcessModel found.')
+
+
+class SendProcess(AbstractProcess):
```
awintel, line 41: Didn't we want to factor such shared collateral out such that many processes can use it? 

If you are the first to need/write it, then start by putting it into a shared place.

awintel: This could even be more than just testing infrastructure and go into the io process lib.

tests/lava/proc/conv/test_models.py
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    return conv, input_shape, output_shape, params
+
+
+class TestConvProcessModels(unittest.TestCase):
+    """Tests for all ProcessModels of Conv"""
+
+    def test_source_sink(self):
+        """Test for source-sink process."""
```
awintel, line 187: More details please: What does this test check? What is its purpose?

If necessary, provide inline comments to explain the intent of what you are doing.

awintel (changes requested):
Looks good. Just some minor comments.

src/lava/proc/conv/process.py
bamsumit: That is one of my bad habits... I usually do that in comments :(

src/lava/proc/conv/process.py
bamsumit: It is needed for/simplifies output shape calculation. At this level I can create a var and access var.init but that would still be a scalar. Correct me if I got it wrong.

src/lava/proc/conv/utils.py
bamsumit: numpy.convolve only works for 1d convolution. Besides we would need scipy for sparse matrix representation at some point.

src/lava/proc/conv/process.py
bamsumit: That's one of my bad habit. I tend to do that in comments.

src/lava/proc/conv/process.py
bamsumit: This simplifies output_shape calculation. Var.init at this point would give the scalar value. Correct me if I am wrong.

src/lava/proc/conv/utils.py
bamsumit: np.convolution only works for 1d data. We would need to use scipy for sparse matrix at some point anyway.

src/lava/proc/conv/process.py
awintel: Yes, Var.init will only hold the initially assigned value.
I would have assumed that your output_shape helper function could simply compute the output shape from 'shape'.

tests/lava/proc/conv/test_models.py
bamsumit: sure I will move them to lava.proc.io

src/lava/proc/conv/models.py
bamsumit: sure, that's for later addition I guess.

src/lava/proc/conv/process.py
bamsumit: It's a local function that gets killed after it servers its purpose in init. So I am not too worried about resources here.

srrisbud: won't it be good to put this in utils?

src/lava/proc/io/sink.py
```diff
@@ -0,0 +1,65 @@
...
+        data = self.a_in.recv()
+        buffer = self.data.shape[-1]
+        self.data[..., (self.current_ts - 1) % buffer] = data
+
+
+@implements(proc=ReceiveProcess, protocol=LoihiProtocol)
+@requires(CPU)
+@tag('floating_pt')
+class PyReceiveModelFloat(AbstractPyReceiveModel):
+    """Float receive process model."""
```
srrisbud, line 54: Preferable description: "Floating point implementation of ..."

src/lava/proc/io/sink.py
```diff
@@ -0,0 +1,65 @@
...
+    a_in: PyInPort = LavaPyType(PyInPort.VEC_DENSE, np.int32, precision=24)
+    data: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=24)
+
+
+@implements(proc=ReceiveProcess, protocol=LoihiProtocol)
+@requires(CPU)
+@tag('fixed_pt')
+class PyReceiveModelFixed(AbstractPyReceiveModel):
+    """Fixed point receive process model."""
+    a_in: PyInPort = LavaPyType(PyInPort.VEC_DENSE, float)
```
srrisbud, line 64: See the previous comment

src/lava/proc/io/source.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        data = kwargs.pop('data')
+        self.data = Var(shape=data.shape, init=data)
+        self.s_out = OutPort(shape=data.shape[:-1])  # last dimension is time
+
+
+class AbstractPySendModel(PyLoihiProcessModel):
+    """Template send process model."""
```
srrisbud, line 34: Not a "Template". "Abstract base.." or "Parent"

src/lava/proc/io/source.py
```diff
@@ -0,0 +1,58 @@
...
+    s_out: PyOutPort = LavaPyType(PyOutPort.VEC_DENSE, float)
+    data: np.ndarray = LavaPyType(np.ndarray, float)
+
+
+@implements(proc=SendProcess, protocol=LoihiProtocol)
+@requires(CPU)
+@tag('fixed_pt')
+class PySendModelFixed(AbstractPySendModel):
+    """Fixed point send process model."""
+    s_out: PyOutPort = LavaPyType(PyOutPort.VEC_DENSE, np.int32, precision=24)
```
srrisbud, line 57: Somewhere in here, justify and/or document the choice of 24-bit precision for the fixed point model. We are doing it to be compatible with Loihi 2 hardware's graded spikes. But the model is just tagged 'fixed_pt', and has no notion of being bit accurate or anything.

tests/lava/proc/conv/test_models.py
```diff
@@ -0,0 +1,204 @@
...
+from lava.magma.core.run_configs import RunConfig
+from lava.magma.core.run_conditions import RunSteps
+from lava.proc.conv.process import Conv
+from lava.proc.conv import utils
+from lava.proc.io.source import SendProcess
+from lava.proc.io.sink import ReceiveProcess
+
+
+verbose = True if (('-v' in sys.argv) or ('--verbose' in sys.argv)) else False
+TORCH_IS_AVAILABLE = utils.TORCH_IS_AVAILABLE
```
srrisbud, line 17: Add ToDo for TF support

srrisbud (changes requested):
There are a few changes needed. Especially the ReceiveProcess has its ProcessModels flipped.

src/lava/proc/io/sink.py
```diff
@@ -0,0 +1,65 @@
...
+from lava.magma.core.sync.protocols.loihi_protocol import LoihiProtocol
+from lava.magma.core.model.py.ports import PyInPort
+from lava.magma.core.model.py.type import LavaPyType
+from lava.magma.core.resources import CPU
+from lava.magma.core.decorator import implements, requires, tag
+from lava.magma.core.model.py.model import PyLoihiProcessModel
+
+
+class ReceiveProcess(AbstractProcess):
+    """Receive process
```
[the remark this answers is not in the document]
awintel: By the way is sink.py the best module name?

We have this dissonance in naming:
- package called 'io' meaning input/output
- module called sink
- class called Receive

input proc <-> sender <-> source
out proc <-> receiver <-> sink

src/lava/proc/conv/utils.py
```diff
@@ -0,0 +1,236 @@
...
+
+import numpy as np
+from scipy import signal
+
+try:
+    import torch
+    import torch.nn.functional as F
+    TORCH_IS_AVAILABLE = True
+except ModuleNotFoundError:
+    TORCH_IS_AVAILABLE = False
```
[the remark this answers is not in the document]
bamsumit: Yeah, torch implementation was just a test. We should definitely do something with tensorflow as well. The reason I didn't do it now is because grouping implementation needs work in tensorflow to align with similar interface. But yes, that's definitely a TODO.

The reason I didn't put the broadcast in utils is it will likely not be useful elsewhere except the initializaion. Plus it will be local here.

src/lava/proc/io/source.py
awintel: Also, as long as you are not using the precision anywhere in your code, there is no real need of specifying it in the LavaPyType.
Otherwise people will ask "what is this for"

awintel (commented):
Just some additional comments

src/lava/proc/conv/utils.py
```diff
@@ -0,0 +1,236 @@
...
+    bits : int
+        number of bits for the variable
+
+    Returns
+    -------
+    same type as x
+        clamped value
+    """
+    base = 1 << bits
+    return (x - base // 2) % base + base // 2  # signed value clamping
```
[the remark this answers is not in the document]
bamsumit: My bad. Should have been 
`(x + base // 2) % base - base // 2`
We want underflow/overflow behavior, isn't it? so np.clip would not work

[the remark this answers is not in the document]
bamsumit: I've fixed it

src/lava/proc/io/sink.py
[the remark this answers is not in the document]
bamsumit: I am open to suggestions on the naming. it could well be `sink.BasicSink` or `reveive.BasicReceive`

src/lava/proc/io/sink.py
```diff
@@ -0,0 +1,65 @@
...
+        buffer = self.data.shape[-1]
+        self.data[..., (self.current_ts - 1) % buffer] = data
+
+
+@implements(proc=ReceiveProcess, protocol=LoihiProtocol)
+@requires(CPU)
+@tag('floating_pt')
+class PyReceiveModelFloat(AbstractPyReceiveModel):
+    """Float receive process model."""
+    a_in: PyInPort = LavaPyType(PyInPort.VEC_DENSE, np.int32, precision=24)
```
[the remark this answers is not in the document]
bamsumit: good catch ;)

src/lava/proc/io/source.py
bamsumit: Actually it would be good to parameterize the precision as an argument to the process definition. But I am not clear how that can be retrieved during Model definition at the moment.

src/lava/proc/io/sink.py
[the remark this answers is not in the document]
awintel: In general I'm voting for:
- package: io
- Different modules/packages for different types of input and output processes. The module/package names should be consistent with the class names found therein. If you call the module sink.py, don't call the process ReceiveProcess. In Particular there is no need to call it 'Process' since it is-a Process.

Before we assign too generic-sounding names for e.g. an Input process, we should think a bit on what type of input processes there will be in the future. You have a very particular type of input process that just injects a preprogrammed array of data over time. How should we call this versus one that loads images/arrays from a file? Gets it from a DVS camera, etc...

src/lava/proc/conv/process.py
bamsumit: The reason I am hesitant to put this to utils is in this form, it uses kwargs from init space. This makes these lines simple
`padding = broadcast_arg('padding', 0)`
Whereas it would be
`padding = utils.broadcast_arg(kwargs.get('padding', 0))`
I am happy to do that as well.

src/lava/proc/io/source.py
awintel: Check the builder. I originally put some code in there that would save the precision as an attribute. But currently somebody has commented it out.

As an alternative, you could simply access the LavaPyType:
MyProcClass.some_class_attr.precision

What's actually missing from the LavaPyType is a specification of whether it's signed or unsigned. I could imagine three ways of doing this:
- Introduce another argument for LavaPyType. But adds extra verbosity.
- Introduce negative and positive precision numbers. Negative meaning signed, Positive meaning unsigned. But could be unusual and confusing.
- Using the data type to infer that like np.unit8 or np.int8. Requires that people think consciously about int and uint which is probably ok.

src/lava/proc/conv/process.py
awintel: Yes could go in utils. 

But forget my original comment. I did not read it carefully. I thought you were broadcasting big data arrays here. But you are actually just building a shape tuple. However, in this case the name is 'broadcast' is misleading because you are not actually broadcasting anything.

tests/lava/proc/conv/test_utils.py
```diff
@@ -0,0 +1,86 @@
...
+    # in this case, the test compares against torch ground truth
+except ModuleNotFoundError:
+    compare = False
+    # in this case, the test only checks for error during
+    # utils.conv calculation
+
+
+class TestConv(unittest.TestCase):
+    def test_conv(self):
+        """Test convolution implementation"""
```
drager-intel, line 23: Do we have any method of testing the behavior of Conv when Torch is not available? I feel as though there should be some basic unit tests with weight mats/connectivity structures of specialized forms to ensure that inputs are being summed properly, even in the absence of Torch.

drager-intel (changes requested):
My only 2 comments are surrounding unit tests. Can we add some unit tests for the Conv Connection behavior that don't rely on Torch?

tests/lava/proc/conv/test_process.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (C) 2021 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+# See: https://spdx.org/licenses/
+import unittest
+import numpy as np
+from lava.proc.conv.process import Conv
+
+
+class TestLIFProcess(unittest.TestCase):
```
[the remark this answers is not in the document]
bamsumit: Good catch!

src/lava/proc/conv/models.py
```diff
@@ -0,0 +1,63 @@
...
+    """Abstract template implemetation of PyConvModel."""
+    s_in = None
+    a_out = None
+    weight = None
+
+    kernel_size: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=8)
+    stride: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=8)
+    padding: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=8)
+    dilation: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=8)
+    groups: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=8)
```
mathisrichter, line 28: Should these be of type np.uint8 rather than np.int32?

src/lava/proc/conv/models.py
```diff
@@ -0,0 +1,63 @@
...
+from lava.magma.core.model.py.type import LavaPyType
+from lava.magma.core.resources import CPU
+from lava.magma.core.decorator import implements, requires, tag
+from lava.magma.core.model.py.model import PyLoihiProcessModel
+from lava.proc.conv.process import Conv
+
+from lava.proc.conv import utils
+
+
+class AbstractPyConvModel(PyLoihiProcessModel):
```
mathisrichter, line 18: Is this an abstract class? If not, please remove the 'Abstract' from the name.

src/lava/proc/conv/models.py
```diff
@@ -0,0 +1,63 @@
...
+    s_in: PyInPort = LavaPyType(PyInPort.VEC_DENSE, bool, precision=1)
+    a_out: PyOutPort = LavaPyType(PyOutPort.VEC_DENSE, float)
+    weight: np.ndarray = LavaPyType(np.ndarray, float)
+
+
+@implements(proc=Conv, protocol=LoihiProtocol)
+@requires(CPU)
+@tag('fixed_pt')
+class PyConvModelBinaryFixed(AbstractPyConvModel):
+    """Binary spike fixed point synapse implementation."""
```
mathisrichter, line 57: This is ambiguous. How about something like "Binary spikes and synapses based on fixed-point values"

src/lava/proc/conv/process.py
```diff
@@ -0,0 +1,90 @@
...
+        kernel_size = kwargs['weight'].shape[1:3]
+        in_channels = input_shape[-1]
+        out_channels = kwargs['weight'].shape[0]
+        padding = utils.get_tuple(kwargs, 'padding', 0)
+        stride = utils.get_tuple(kwargs, 'stride', 1)
+        dilation = utils.get_tuple(kwargs, 'dilation', 1)
+        groups = kwargs.get('groups', 1)
+
+        if len(input_shape) != 3:
+            raise Exception(
```
mathisrichter, line 54: Raise more specific exception type here and below

src/lava/proc/conv/utils.py
```diff
@@ -0,0 +1,274 @@
...
+        )
+
+    return x_out, y_out, out_channels
+
+
+# TODO: This might be needed later. The reverse mapping is ambiguous though
+# def input_shape(
+#     output_shape, in_channels, kernel_size, stride, padding, dilation
+# ):
+#     pass
```
mathisrichter, line 150: Remove for release?

src/lava/proc/conv/utils.py
```diff
@@ -0,0 +1,274 @@
...
+    groups : int
+        number of convolution groups.
+
+    Returns
+    -------
+    3 dimensional np array
+        convolution output
+    """
+    if TORCH_IS_AVAILABLE:
+        # with torch.no_grad():  # this seems to cause problems
```
mathisrichter, line 179: Remove for release

src/lava/proc/conv/utils.py
```diff
@@ -0,0 +1,274 @@
...
+        output = F.conv2d(
+            torch.unsqueeze(  # torch expects a batch dimension NCHW
+                torch.FloatTensor(input.transpose([2, 1, 0])),
+                dim=0,
+            ),
+            torch.FloatTensor(
+                # torch acutally does correlation
```
mathisrichter, line 186: typo in actually

src/lava/proc/io/sink.py
```diff
@@ -0,0 +1,66 @@
...
+        super().__init__(**kwargs)
+        shape = kwargs.get('shape', (1,))
+        buffer = kwargs.get('buffer')
+        self.shape = shape
+        self.a_in = InPort(shape=shape)
+        buffer_shape = shape + (buffer,)
+        self.data = Var(shape=buffer_shape, init=np.zeros(buffer_shape))
+
+
+class AbstractPyReceiveModel(PyLoihiProcessModel):
```
mathisrichter, line 39: See above: Is this class abstract?

tests/lava/proc/conv/test_models.py
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    return conv, input_shape, output_shape, params
+
+
+class TestConvProcessModels(unittest.TestCase):
+    """Tests for all ProcessModels of Conv"""
+
+    def test_source_sink(self):
+        """Test whatever is being sent form source is received at sink."""
```
mathisrichter, line 86: form -> from

tests/lava/proc/conv/test_models.py
```diff
@@ -0,0 +1,204 @@
...
+        num_steps = 10
+        shape = np.random.randint([128, 128, 16]) + 1
+        input = np.random.randint(
+            256,
+            size=(shape).tolist() + [num_steps]
+        ) - 128
```
mathisrichter, line 92: Please do not use random numbers in unit tests as they may introduce non-deterministic behavior. If you want to test ranges of values, do so explicitly. If you just need some data, hard-code it or use randomly generated numbers with a fixed random seed.

tests/lava/proc/conv/test_models.py
```diff
@@ -0,0 +1,204 @@
...
+            f'{input[output!=input] =}'
+        )
+
+    def test_conv_float(self):
+        """Test for float conv process."""
+        num_steps = 10
+        utils.TORCH_IS_AVAILABLE = False
+
+        conv, input_shape, output_shape, params = setup_conv()
+        input = np.random.random(input_shape + (num_steps,))
```
mathisrichter, line 119: See above; fix throughout

tests/lava/proc/conv/test_utils.py
```diff
@@ -0,0 +1,88 @@
...
+            ]
+            weights = np.random.randint(256, size=weight_dims) - 128
+            input = np.random.random(
+                (
+                    # input needs to be a certain size
+                    # to make sure the output dimension is never negative
+                    np.random.randint([128, 128]) + kernel_size * dilation
+                ).tolist()
+                + [in_channels]
+            )
```
mathisrichter, line 46: See comment above about random numbers

tests/lava/proc/io/test_source_sink.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class TestSendReceive(unittest.TestCase):
+    """Tests for all SendProces and ReceiveProcess."""
+
+    def test_source_sink(self):
+        """Test whatever is being sent form source is received at sink."""
+        num_steps = 10
+        shape = np.random.randint([128, 128, 16]) + 1
+        input = np.random.randint(256, size=(shape).tolist() + [num_steps])
```
mathisrichter, line 34: See above

mathisrichter (changes requested):
Looks great, Sumit!
- Not sure whether this is a general convention but elsewhere we use type hints in class, method, and function signatures. Maybe add an issue for that and do it later.

src/lava/proc/conv/models.py
```diff
@@ -0,0 +1,63 @@
...
+from lava.magma.core.resources import CPU
+from lava.magma.core.decorator import implements, requires, tag
+from lava.magma.core.model.py.model import PyLoihiProcessModel
+from lava.proc.conv.process import Conv
+
+from lava.proc.conv import utils
+
+
+class AbstractPyConvModel(PyLoihiProcessModel):
+    """Abstract template implemetation of PyConvModel."""
```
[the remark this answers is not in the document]
bamsumit: Are you talking about these lines?
```
class AbstractPyConvModel(PyLoihiProcessModel):
    """Abstract template implementation of PyConvModel."""
    s_in = None
    a_out = None
    weight = None
```

src/lava/proc/conv/models.py
bamsumit: Yes it's an abstract class

mathisrichter: I was rather asking how it is an abstract class as I didn't see any methods that needed to be implemented by subclasses? Am I missing something?

src/lava/proc/conv/utils.py
bamsumit: If we remove for release, my concern is that it might get lost along the way later.

src/lava/proc/conv/utils.py
bamsumit: I am pretty sure I'll try and do `with torch.no_grad()` later if I don't see the comment. Because that would be memory efficient in principle... haha

src/lava/proc/io/sink.py
bamsumit: Yes

src/lava/proc/conv/models.py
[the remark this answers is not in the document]
mathisrichter: > **Type hints:** Are you talking about these lines?

Not specifically - I was referring to the entire code. Any method signature should have type hints - I believe.
Something like
```python
def my_method(shape: typing.Tuple[int, ...], amplitude: int) -> np.ndarray:
    pass
```

src/lava/proc/conv/utils.py
mathisrichter: You could create an issue instead

tests/lava/proc/conv/test_models.py
bamsumit: I would like to use random settings to test them otherwise grid search space is quite large. Typically I use initialize the random seed and print out the seed when there is error so that one can debug it. Here I am printing out the dimension settings instead.

src/lava/proc/conv/models.py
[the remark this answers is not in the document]
bamsumit: I am not sure what convention we have settled on. I looked around other modules and and didn't see it strictly followed.

tests/lava/proc/conv/test_models.py
mathisrichter: What if all tests pass on your system and the test fails in CI/CD, or in production? I'd suggest to rather test all qualitatively different settings (even if it takes long) than to do random tests.

What do other people think? @mgkwill @awintel

src/lava/proc/conv/models.py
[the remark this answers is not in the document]
mathisrichter: I don't know either. @awintel didn't complain earlier so maybe it's left to personal choice. But the code would look more homogeneous if we had a convention.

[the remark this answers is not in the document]
awintel: Yes, I totally forgot that. Yes, all methods should have type annotations for inputs and return arguments.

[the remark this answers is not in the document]
awintel: This should be done consistently throughout the code base. If you see otherwise, please point it out in the future.

tests/lava/proc/conv/test_models.py
awintel: If you need some randomness, you could still seed you RNG but then have an outer loop running multiple instances of your base test with different random numbers. This should give you broad but still deterministic coverage.

src/lava/proc/conv/models.py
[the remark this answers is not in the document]
bamsumit: Roger that. We should then enable `flake8-annotations` or something similar then rather than going in and checking manually.

mgkwill:
> Starting pull request

This is probably clear to some due to internal conversations but the title is not descriptive and I don't see an issue for this pull.

Can we please follow the process, I.E. open an issue that explains the change and have at least a minimal explanation of the work in the pull request?

tests/lava/proc/conv/test_models.py
bamsumit: Yeah, I am doing 10 random runs of same test in a loop/across time at the moment.

tests/lava/proc/conv/test_models.py
```diff
@@ -15,23 +17,32 @@
 
 verbose = True if (('-v' in sys.argv) or ('--verbose' in sys.argv)) else False
 TORCH_IS_AVAILABLE = utils.TORCH_IS_AVAILABLE
+seed = np.random.randint(1000)
+np.random.seed(seed)
```
mathisrichter, line 21: Maybe I am missing something but setting the random seed to a randomly generated number does not make unit tests predictable. I think Andreas as suggesting to set the seed to a fixed value, or to run the test in a loop, where each iteration sets the random seed to a (different) fixed value.

Everything else looks great.

mathisrichter (changes requested):

tests/lava/proc/conv/test_models.py
awintel: Indeed ;-)

bamsumit: What I want in the unittest is to have unpredictability. That is the point of having the argument dimensions randomly generated. If we set predictable 10 seed values, then the test would only test for that combination of 10 settings. What if there is issue in any case that is not covered in the 10 settings? I am trying to do this because we would not want to slow down the build.
With unpredictablity, every time unittest is run, we are testing different setting and the testing confidence grows over time. Once such failure seeds are found, those would then be used to add more custom test for that specific setting. I think this is called monkey testing.
The logic above is just keeping track of seed so that was used in the experiment which would show up once the test fails and we can use the seed to replicate the exact setting for with the test failed and investigate more.

bamsumit: Perhaps we can use fixed seeds here and design an elaborate test for nightly/weekly test?

bamsumit: Okay, I googled around for and against random testing. It's not a good idea to randomize CI/CD test esp for large projects. My bad. I'll switch to fixed seeds.

mathisrichter: No worries, Sumit. I completely understand your reasoning and I would often like to do the same thing.

We could think about setting up a separate infrastructure for large-scale tests that need time to run, for instance for complete grid searches. These tests would then be decoupled from the regular CI/CD and would only be run from time to time on our servers. Something like that.
Maybe there is an established best practice for that @mgkwill ?

mathisrichter (approved):

awintel (approved):
Nothing major except for missing indentation of function arguments that start in a new line. That makes code not very readable. 
Could be fixed later but we should not leave it like that.

drager-intel (approved):
Thanks for adding a known input-output unit test of conv behavior for users w/out Torch installed! Looks good!

srrisbud (approved):

PR #77, opened 2021-11-19
PyInPort: probe() implementation

gkarray:
Implementation of `probe` method for PyInPorts.

Added : 

- Implementation of `probe` method for PyInPorts.
- Two simple unit tests for `probe` method for `PyInPortVectorDense` only.


awintel:
I suggest to always include a few more people than just the two minimum required number of reviewers.

src/lava/magma/core/model/py/ports.py
```diff
@@ -48,7 +48,11 @@ def peek(self):
         pass
 
     def probe(self) -> bool:
-        pass
+        return ft.reduce(
+            lambda acc, csp_port: acc and csp_port.probe(),
+            self._csp_recv_ports,
+            True,
+        )
```
mathisrichter, line 55: I don't understand what this does. Please add a descriptive comment.

tests/lava/magma/core/model/py/test_ports.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class TestPyPorts(unittest.TestCase):
+    def test_py_in_port_vector_dense_probe_single_csp_recv_port(self):
+        """Tests PyInPortVectorDense probe method for single recv_csp_port"""
+        smm = SharedMemoryManager()
+        try:
+            smm.start()
+
+            data = np.random.randint(5, size=(2, 4))
```
mathisrichter, line 42: Is it important that this is randomized? Random numbers in unit tests are problematic because the tests become non-deterministic and may sometimes pass or fail.
It seems, here, that you could just use a hard-coded values. Otherwise fix the random-seed.

tests/lava/magma/core/model/py/test_ports.py
```diff
@@ -0,0 +1,127 @@
...
+        finally:
+            smm.shutdown()
+
+    def test_py_in_port_vector_dense_probe_multi_csp_recv_port(self):
+        """Tests PyInPortVectorDense probe method for multiple recv_csp_port"""
+        smm = SharedMemoryManager()
+        try:
+            smm.start()
+
+            data = np.random.randint(5, size=(2, 4))
```
mathisrichter, line 82: See above

tests/lava/magma/core/model/py/test_ports.py
```diff
@@ -0,0 +1,127 @@
...
+            _ = recv_py_port.recv()
+            # Probe PyInPort
+            probe_value = recv_py_port.probe()
+
+            # probe_value should be False since PyInPort's buffer was emptied
+            self.assertFalse(probe_value)
+        finally:
+            smm.shutdown()
+
+    def test_py_in_port_vector_dense_probe_multi_csp_recv_port(self):
```
mathisrichter, line 76: The two unit tests are almost identical. Could you refactor them to avoid code duplication? Maybe sync with @mgkwill on that, he has the same issue.

mathisrichter (changes requested):

tests/lava/magma/core/model/py/test_ports.py
gkarray: You're right, no need to use random numbers here.
This has been addressed in latest commit.

tests/lava/magma/core/model/py/test_ports.py
gkarray: This has been addressed in latest commit.

src/lava/magma/core/model/py/ports.py
gkarray: Added a descriptive comment in latest commit.

tests/lava/magma/core/model/py/test_ports.py
gkarray: Refactored unit tests into a single test as there isn't a particular reason for testing probe feature in single and multi csp recv ports individually.

Plus, refactored the test "routine" to be easily usable for multiple PyInPort implementations.

src/lava/magma/core/model/py/ports.py
```diff
@@ -48,6 +48,13 @@ def peek(self):
         pass
 
     def probe(self) -> bool:
+        """Executes probe method of all csp ports and accumulates the returned
+        bool values with and operation. The accumulator acc is initialized to
```
mathisrichter, line 52: Capitalize "and" into "AND" to distinguish it from the regular word.

mathisrichter (changes requested):

src/lava/magma/core/model/py/ports.py
gkarray: Resolved!

src/lava/magma/core/model/py/ports.py
```diff
@@ -48,7 +48,18 @@ def peek(self):
...
+        """Executes probe method of all csp ports and accumulates the returned
+        bool values with AND operation. The accumulator acc is initialized to
+        True.
+
+        Returns the accumulated bool value.
+        """
+        # Returns True only when probe returns True for all _csp_recv_ports.
+        return ft.reduce(
+            lambda acc, csp_port: acc and csp_port.probe(),
+            self._csp_recv_ports,
```
PhilippPlank, line 60: Maybe use the property csp_ports instead.

PhilippPlank (approved):
Looks good to me for now. At some point we might want to enable InPorts which do not expect data on every csp channel to not block (current behavior, this comes from the fact that previously each InPort had exactly one csp channel), but this is for later I guess.

src/lava/magma/core/model/py/ports.py
gkarray: I was actually going to, but then saw different other methods in the same file using _csp_recv_ports and _csp_send_ports instead of csp_ports property.
Left _csp_recv_ports usage for homogeneity. Should I change it ?

mathisrichter (approved):

mgkwill (approved):

awintel:
If an InPort has multiple inputs, it expects inputs from all of them. The only way to not block on some input is to receive a flush.

Issue #86, opened 2021-11-21
Remove hacks and refactor Runtime and RuntimeService/Synchronizer


awintel:
The Runtime and RuntimeService do currently contain a number of hacks that need to be cleaned up:

- The Runtime has a 'current_ts' attribute. But the Runtime must not be aware of algorithmic time steps. That's up to the Synchronizer to and ProcessModels to implement.
- The Runtime currently misuses the runtime_to_service_cmd channel to send num_steps.
- The RuntimeService should be split into a RuntimeService and a pure Synchronizer. A universal RuntimeService should service the commands received by the Runtime and then delegate these commands to its Synchronizer which can be different for each RuntimeService.
- The RunContinuous RunCondition is not supported.
- The Asynch Synchronizer is currently not thought through. No clear contract between Process and Synchronizer is defined, hence there are numerous ways this can deadlock.

Side question: Should we use asyncio in PyProcModels and PyRuntimeService instead of busy waiting on channels? Will this be more efficient?

awintel:
I started addressing some of these issues.

It should be possible to create a completely generic PyRuntimeService. The responsibility of this PyRuntimeService is merely to communicate RUN, PAUSE, STOP commands between Runtime, and PyProcModels as well as any SET/GET requests. These are generic communication capabilities that are independent of the specific SyncProtocol.

The current LoihiPyRuntimeService conceptually covers these capabilities and can be refactored to become this generic PyRuntimeService except for its run() method. 

Only the part of LoihiPyRuntimeService dealing with the different SPK, MGMT, ... phases are specific to a particular SyncProtocol (i.e. the LoihiSyncProtocol) and should therefore be factored out and provided as an object to a refactored PyRuntimeService.

While beginning to refactor LoihiPyRuntimeService, I began to set up a cleaner state machine structure:
- The PyRuntimeService should have a dictionary mapping commands to function objects, aka 'command handlers'.
- Each command handler (i.e. what to do when RUN, PAUSE, ... is received) is a separate method of the PyRuntimeService.
- The core _run(..) method merely iterates over all channels, waits for a command and when it receives one, get and executes the appropriate command handler from the dictionary.
- The basic PyRuntimeService only has RUN, PAUSE, STOP, GET, SET in its command handler dictionary.
[...]

PR #87, opened 2021-11-21
Performance improvements

harryliu-intel:
A few performance improvements:
- replace probe() plus busy waiting with select()
- use a specialized function to compare messages of enum constants
- use multiprocessing semaphore to avoid overhead of pipe

### DIFF
--- a/src/lava/magma/compiler/channels/pypychannel.py
+++ b/src/lava/magma/compiler/channels/pypychannel.py
@@ -3,12 +3,12 @@
 # See: https://spdx.org/licenses/
 import typing as ty
 from queue import Queue, Empty
-from threading import Thread
+from threading import BoundedSemaphore, Condition, Thread
 from time import time
 from dataclasses import dataclass
 
 import numpy as np
-from multiprocessing import Pipe, BoundedSemaphore
+from multiprocessing import Pipe
 
 from lava.magma.compiler.channels.interfaces import (
     Channel,
@@ -57,6 +57,7 @@ class CspSendPort(AbstractCspSendPort):
         self._done = False
         self._array = []
         self._semaphore = None
+        self.observer = None
         self.thread = None
 
     @property
@@ -97,7 +98,10 @@ class CspSendPort(AbstractCspSendPort):
         try:
             while not self._done:
                 self._ack.recv_bytes(0)
+                not_full = self.probe()
                 self._semaphore.release()
+                if self.observer and not not_full:
+                    self.observer()
         except EOFError:
             pass
 
@@ -190,6 +194,7 @@ class CspRecvPort(AbstractCspRecvPort):
         self._done = False
         self._array = []
         self._queue = None
+        self.observer = None
         self.thread = None
 
     @property
@@ -230,7 +235,10 @@ class CspRecvPort(AbstractCspRecvPort):
         try:
             while not self._done:
                 self._req.recv_bytes(0)
+                not_empty = self.probe()
                 self._queue.put_nowait(0)
+                if self.observer and not not_empty:
+                    self.observer()
         except EOFError:
             pass
 
@@ -263,6 +271,43 @@ class CspRecvPort(AbstractCspRecvPort):
 
     def join(self):
         self._done = True
+
+
+class CspSelector:
+    """
+    Utility class to allow waiting for multiple channels to become ready
+    """
+
+    def __init__(self):
+        """Instantiates CspSelector object and class attributes"""
+        self._cv = Condition()
+
+    def _changed(self):
+        with self._cv:
+            self._cv.notify_all()
+
+    def _set_observer(self, channel_actions, observer):
+        for channel, _ in channel_actions:
+            channel.observer = observer
+
+    def select(
+        self,
+        *args: ty.Tuple[ty.Union[CspSendPort, CspRecvPort],
+                        ty.Callable[[], ty.Any]
+                        ]
+    ):
+        """
+        Wait for any channel to become ready, then execute the corresponding
+        callable and return the result.
+        """
+        with self._cv:
+            self._set_observer(args, self._changed)
+            while True:
+                for channel, action in args:
+                    if channel.probe():
+                        self._set_observer(args, None)
+                        return action()
+                self._cv.wait()
 
 
 class PyPyChannel(Channel):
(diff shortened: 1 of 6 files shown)


awintel:
Oh, oh... that means work for me. Not knowing what you would be doing. I have also refactored the LoihiPyProcessModel by pushing a lot of the generic MGMT command handling into the AbstractPyProcModel since this is not Loihi specific and should also be the same for AsyncProcs e.g. on a CPU.

For that purpose I introduced a message handler based on a dictionary mapping from commands to function objects/message handlers. The AbstractPyProcModel sets up a basic dictionary with RUN, PAUSE, STOP and respective message handlers. Sub classes like LoihiPyProcModel would then just extend the set of handlers. This made the code much cleaner.

Will first finish my stuff then revisit your changes.

What in all of your changes was the key to making it more performant and by how much?

mgkwill:
I like the enum_equal fucntion and using multiprocessing semaphore also seems like a good idea. 

This change also has the likelihood of affect current inflight Pyport changes but will adapt as necessary.

harryliu-intel:
Sorry, Andreas, I didn't know you'd be working in the same area.  The most significant speed up comes from the select() commit, but I think each of the other 2 commits also makes a measurable difference.

awintel:
No worries. Not your fault ;-)

src/lava/magma/core/model/py/model.py
```diff
@@ -148,37 +147,44 @@ def run(self):
...
+
+            channel_actions = [(self.service_to_process_cmd, lambda: 'cmd')]
+            if enum_equal(phase, PyLoihiProcessModel.Phase.PRE_MGMT) or \
+                    enum_equal(phase, PyLoihiProcessModel.Phase.POST_MGMT):
+                for var_port in self.var_ports:
+                    for csp_port in var_port.csp_ports:
+                        if isinstance(csp_port, CspRecvPort):
+                            channel_actions.append((csp_port, lambda: var_port))
+            elif enum_equal(phase, PyLoihiProcessModel.Phase.HOST):
+                channel_actions.append((self.service_to_process_req,
```
awintel, line 170: From a type perspective it is a bit inconsistent to give either a string or a port back as an action.

src/lava/magma/compiler/channels/pypychannel.py
```diff
@@ -265,6 +273,43 @@ def join(self):
...
+class CspSelector:
+    """
+    Utility class to allow waiting for multiple channels to become ready
+    """
+
+    def __init__(self):
+        """Instantiates CspSelector object and class attributes"""
+        self._cv = Condition()
+
+    def _changed(self, channel):
```
awintel, line 285: channel is not used

src/lava/magma/compiler/channels/pypychannel.py
```diff
@@ -97,7 +98,10 @@ def _ack_callback(self):
         try:
             while not self._done:
                 self._ack.recv_bytes(0)
+                not_full = self.probe()
                 self._semaphore.release()
+                if self.observer and not not_full:
+                    self.observer(self)
```
awintel, line 104: CspSelector._changed(..) does not need its input argument (self in this case).

src/lava/magma/compiler/channels/pypychannel.py
```diff
@@ -265,6 +273,43 @@ def join(self):
...
+        *args: ty.Tuple[ty.Union[CspSendPort, CspRecvPort],
+                        ty.Callable[[], ty.Any]
+                        ]
+    ):
+        """
+        Wait for any channel to become ready, then execute the corresponding
+        callable and return the result.
+        """
+        with self._cv:
+            self._set_observer(args, self._changed)
```
awintel, line 304: Why do we have to set and unset this all the time?

src/lava/magma/compiler/channels/pypychannel.py
```diff
@@ -230,7 +235,10 @@ def _req_callback(self):
         try:
             while not self._done:
                 self._req.recv_bytes(0)
+                not_empty = self.probe()
```
awintel, line 238: This loop is still busy waiting, isn't it? Is this a performance concern? Could this also be improved in performance by using a condition?

awintel (commented):
Took me a bit to understand what's happening here...

My interpretation is the following:
- You introduced a CspSelector which essentially has a threading.Condition object. 
- CspSelector.select is like a message handler of its own. It is given a list of channels and actions to perform when there is a token on any of the given channels.
- The select method passes the central Condition of the CspSelector to each channel.
- The CspSelector then simply waits for any channel to trigger the condition. This Condition.wait() is probably implemented more efficiently than busy waiting.
- Once any channel has a new token it sends a notification via the threading.Condition to all observers. This unblocks the wait() inside the CspSelector.select() method. We iterate once over all all channels to see which has as token and then call the corresponding action/message handler.

Overall this looks good!
I don't mind if we pull this into the current release if this all works but going forward I hope we can simplify the overall design of the PyRuntimeService and PyProcessModel as outlined in the other git issue:
https://github.com/lava-nc/lava/issues/86

If possible, we should treat all channels equally, have one selector that listens on all available channels and calls an appropriate action when any becomes active. The action could just be a function pointer instead of a string in some cases or  port in some other case.

src/lava/magma/compiler/channels/pypychannel.py
harryliu-intel: In an earlier version, all channels notified, and _changed() determined if the source channel is part of the current set being selected on.  (This is also the reason for the now unused channel argument, which I didn't bother to remove.)  I thought doing it this way was cheaper and cleaner.

src/lava/magma/compiler/channels/pypychannel.py
harryliu-intel: There no busy waiting here, because self._req.recv_bytes() would have blocked.

harryliu-intel (commented):
> The action could just be a function pointer instead of a string in some cases or port in some
> other case.

I agree but I avoided that because it would have required larger structural changes that might make this commit more difficult to understand and review.  I would suggest the original authors come back and refactor it.

PhilippPlank (approved):
Overall I think we should merge it, since it improves performance and does not break our unit tests. We still need to refactor our runtime/runtime_service interactions, but one step after the other.

mgkwill (approved):

Issue #12, opened 2021-10-28
Refactor lava messaging and implement remaining PyPorts


mgkwill:
As a user I'd like to be able to use different ports based on sparsity using vectors and scalars. 

Implied ports:
- PyInPortVectorSparse
- PyInPortScalarDense
- PyInPortScalarSparse
- PyOutPortVectorSparse
- PyOutPortScalarDense
- PyOutPortScalarSparse

Retrofit to work with other input types/shapes:
- PyInPortVectorDense

https://github.com/lava-nc/lava/blob/9d80ea15af495b87df7f4b63e83531bdaf7d41f9/lava/magma/core/model/py/ports.py#L53-L75

Ensure that ports process data correctly when receiving from the matrix:

| | Vector | Vector | Scalar | Scalar |
| -----------| ------------- | -------- | ------- | ------- |
| -----------| -Receive- Dense | -Receive- Sparse | -Receive- Dense | -Receive- Sparse |
| Vec -Send- Dense  | ------- | ------- | ------- | ------- |
| Vec -Send- Sparse | -------  | ------- | ------- | ------- |
| Sca -Send- Dense  | ------- | ------- | ------- | ------- |
| Sca -Send- Sparse | -------  | ------- | ------- | ------- |

joyeshmishra:
Closing this issue as discussed with Yash and Marcus. The necessary protocols implemented between Python Processes are just VecDense <-> VecDense and VecSparse <-> VecSparse. Python to C Processes can communication additionally for VecDense <-> SeqDense and VecSparse <-> SeqSparse. Other communication protocols are not supported due to lack of use cases.

PR #92, opened 2021-11-22
Refactor lava messaging and add pyports implementation

mgkwill:
Implements the remaining pyports. See issue #12

### DIFF
--- a/tests/lava/magma/core/model/py/test_ports.py
+++ b/tests/lava/magma/core/model/py/test_ports.py
@@ -5,14 +5,25 @@ import unittest
 import time
 from multiprocessing.managers import SharedMemoryManager
 import numpy as np
+import re
+import scipy.sparse as sparse
+from scipy.sparse.csr import csr_matrix
+
 
 from lava.magma.compiler.channels.interfaces import AbstractCspSendPort, \
     AbstractCspRecvPort
-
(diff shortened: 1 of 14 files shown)


mgkwill:
Perhaps we can start by reviewing the unit tests.

I'm reworking the pyports implementation now. Should be somewhat quick but I found some issues with my first draft.

tests/lava/magma/core/model/py/test_ports.py
```diff
@@ -32,7 +73,94 @@ def get_channel(smm, data, size, name="test_channel") -> PyPyChannel:
...
+                send_py_port.start()
+
+                if ('Sparse' in port_combo[0]):
+                    send_py_port.send(data.toarray(), data.indices)
+                else:
+                    send_py_port.send(data)
+
+                if ('Sparse' in port_combo[1]):
+                    recv_data, recv_idx = recv_py_port.recv()
+                    data = csr_matrix(data)
```
mgkwill, line 129: data here could be a csr_matrix already, need to check before converting.

awintel (commented):
You might want to use itertools.combinations in PortCombos instead of manually writing out all combinations.

Instead of such hacks you might want to just store the data in a dict?
`data = getattr(port_combos, 'data_' + str(port_combo[0]))`

Just another reminder to gate prints by a if statements if you want to keep them around such that unit tests are silent.

Issue #90, opened 2021-11-22
Example code not working


matham:
I'm following the first example code from the readme and I'm running into issues.

Here's the traces:

```py
Python 3.10.0 (tags/v3.10.0:b494f59, Oct  4 2021, 19:00:18) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from lava.proc.dense.process import Dense
>>> from lava.proc.lif.process import LIF
>>>
>>> lif1 = LIF()
>>> dense = Dense()
>>> lif2 = LIF()
>>>
>>> # Connect processes via their directional input and output ports
>>> lif1.out_ports.s_out.connect(dense.in_ports.s_in)
>>> dense.out_ports.a_out.connect(lif2.in_ports.a_in)
>>> from lava.magma.core.run_conditions import RunSteps
>>> from lava.magma.core.run_configs import RunConfig
>>> lif1.run(condition=RunSteps(num_steps=10), run_cfg=SimpleRunConfig(sync_domains=[]))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'SimpleRunConfig' is not defined
```

So then I change `SimpleRunConfig(sync_domains=[])` to `RunConfig()` and get the following:

```py
>>> lif1.run(condition=RunSteps(num_steps=10), run_cfg=RunConfig())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "G:\Python\spnet\lava\src\lava\magma\core\process\process.py", line 394, in run
    executable = self.compile(run_cfg)
  File "G:\Python\spnet\lava\src\lava\magma\core\process\process.py", line 345, in compile
    return compiler.compile(self, run_cfg)
  File "G:\Python\spnet\lava\src\lava\magma\compiler\compiler.py", line 1010, in compile
    proc_map = self._map_proc_to_model(procs, run_cfg)
  File "G:\Python\spnet\lava\src\lava\magma\compiler\compiler.py", line 228, in _map_proc_to_model
    if issubclass(model_cls, AbstractSubProcessModel):
  File "C:\Users\Matthew Einhorn\AppData\Local\Programs\Python\Python310\lib\abc.py", line 123, in __subclasscheck__
    return _abc_subclasscheck(cls, subclass)
TypeError: issubclass() arg 1 must be a class
```

A second issue, is that if I then try again the same command, I get this:
```py
[...]

PR #94, opened 2021-11-22
Fix README.md Example Code

mgkwill:
Fix for issue #90


README.md
```diff
@@ -194,6 +194,24 @@ dense.out_ports.a_out.connect(lif2.in_ports.a_in)
 # Execute process lif1 and all processes connected to it for fixed number of steps
 from lava.magma.core.run_conditions import RunSteps
 from lava.magma.core.run_configs import RunConfig
+
+from lava.magma.core.run_conditions import RunSteps
+from lava.magma.core.run_configs import RunConfig
+
+class SimpleRunConfig(RunConfig):
```
awintel, line 198: Can't we just use a predefined RunConfig rather than making this example more complicated? Just import Loihi1SimConfig
This is not so helpful without explanation. 

I also think the sync_domains is not a mandatory input. For simplicity I would leave that out.

awintel (changes requested):
Try to make it simpler.

README.md
mgkwill: `custom_sync_domains` is mandatory because a list is expected but the default is `None`, thus it will through an error if a  list is not provided.

https://github.com/lava-nc/lava/blob/81336e63783edf6b27f2f019797728b208458cb8/src/lava/magma/core/run_configs.py#L75-L84

This is a bug and should be fixed to `[]` in another issue.

mgkwill (commented):
Worked in jupyter notebook. LGTM.

PR #96, opened 2021-11-23
Added empty list attribute `tags` to `AbstractProcessModel`

srrisbud:
The attribute `tags` will be `setattr`ed by `@tag` decorator. But an empty one is needed to keep `@tag` decorator optional.

Signed-off-by: Risbud, Sumedh <sumedh.risbud@intel.com>


src/lava/magma/core/model/model.py
```diff
@@ -50,6 +50,7 @@ class level attributes with the same name if they exist.
     implements_process: ty.Optional[ty.Type[AbstractProcess]] = None
     implements_protocol: ty.Optional[ty.Type[AbstractSyncProtocol]] = None
     required_resources: ty.List[ty.Type[AbstractResource]] = []
+    tags: ty.List[str] = []
```
awintel, line 53: You might want to update __repr__

awintel (changes requested):
Did you not check previously whether a tag was defined or not, throwing an appropriate error?
If so, this can probably be removed now because there will always be 'tags'.

ashishrao7 (changes requested):
The error message about requiring tags is gone however I run into a new error namely,
"TypeError:issubclass() arg 1 must be a class"
image

src/lava/magma/core/model/model.py
srrisbud: This is done.

srrisbud:
> The error message about requiring tags is gone however I run into a new error namely, "TypeError:issubclass() arg 1 must be a class" image

This is caused when the RunConfig cannot find an appropriate ProcessModel. It is independent of this pull request.

PhilippPlank:
> > The error message about requiring tags is gone however I run into a new error namely, "TypeError:issubclass() arg 1 must be a class" image
> 
> This is caused when the RunConfig cannot find an appropriate ProcessModel. It is independent of this pull request.

Could you still add a meaningful error message if this is the case?

awintel:
I don't know how this arises but SubProcModels should not have any 'tags', 'requires' or 'implements' decorators. 

Something must catch this problem much earlier. The compiler should even get into the situation where model_cls is not a class.

src/lava/magma/core/run_configs.py
```diff
@@ -67,42 +67,60 @@ def select(self,
...
-    SubProcessModel for a Hierarchical Process or else a PyProcessModel for a
-    standard Process. The appropriate PyProcessModel is selected based @tag(
-    'floating_pt') or @tag('fixed_pt'), for floating point precision or Loihi
-    bit-accurate fixed point precision respectively"""
+    `SubProcessModel` for a hierarchical Process or else a `PyProcessModel`
+    for a standard Process.
 
-    def __init__(self, custom_sync_domains=None, select_tag='floating_pt',
-                 select_sub_proc_model=False):
+    If any `SubProcessModel`s are available, then they will be returned first.
```
elvinhajizada, line 75: A bit below, you say that priority is given first exception_proc_model_map. I think this should be also clearified here in docstring

src/lava/magma/core/run_configs.py
```diff
@@ -67,42 +67,60 @@ def select(self,
...
-            elif py_proc_model:
-                if self.select_tag in py_proc_model.tags:
-                    return py_proc_model
-            else:
-                raise AssertionError("No legal ProcessModel found.")
+            # Priority given to SubProcessModels
+            if issubclass(pm, AbstractSubProcessModel) and \
+                    self.select_sub_proc_model:
+                return pm
+            elif issubclass(pm, AbstractPyProcessModel):
```
elvinhajizada, line 113: Is not checking issubclass(pm, AbstractSubProcessModel) duplicated? You could just check once, then do 
`if self.select_sub_proc_model:
    return pm
else: # case of choosing PyProcessModel
  `

src/lava/magma/core/run_configs.py
```diff
@@ -67,42 +67,60 @@ def select(self,
...
+
+    In case of `PyProcessModel`s, an appropriate `PyProcessModel` is selected
+    based on the tags associated with it. Tags are set using the `@tag`
+    decorator. If no tags are set for `ProcessModel`s and no tag is selected,
+    then the first matching `ProcessModel` will be returned.
+
+    If any exceptions are needed, they can be specified in
+    `exception_proc_model_map` dictionary as `{Process: ProcessModel}`
+    key-value pairs. These explicit specifications are given
+    precedence over the subsequent logic for `ProcessModel` selection"""
```
elvinhajizada, line 85: I think, adding method descriptions would be helpful, e.g. for select(..)

src/lava/magma/core/run_configs.py
```diff
@@ -67,42 +67,60 @@ def select(self,
...
+            # Priority given to SubProcessModels
+            if issubclass(pm, AbstractSubProcessModel) and \
+                    self.select_sub_proc_model:
+                return pm
+            elif issubclass(pm, AbstractPyProcessModel):
+                # If ProcessModel has tags AND user asked for a specific tag
+                if len(pm.tags) > 0 and (self.select_tag in pm.tags):
+                    return pm
+                # If ProcessModel has no tags and no one asked for a tag anyway
+                elif len(pm.tags) == 0 and not self.select_tag:
```
elvinhajizada, line 118: what about pm has tags, user asked for a tag, but pm does not have that specific tag?

ashishrao7 (approved):
Lava-optimization runs without problems with the latest changes

src/lava/magma/core/run_configs.py
```diff
@@ -67,42 +67,60 @@ def select(self,
...
+                    return pm
+        # We could not find any ProcessModel in the end:
+        raise AssertionError(f"No ProcessModel could be selected with "
+                             f"tag '{self.select_tag}' for Process "
+                             f"{proc.name}::{proc.__class__.__qualname__}.")
```
elvinhajizada, line 124: I think there should be more than one assertionError to notify user exactly what has gone wrong, foe example in above for loop, while checking different conditions

elvinhajizada (changes requested):
I added few comments and request few changes

src/lava/magma/core/run_configs.py
srrisbud: Rearranged the docstring to reflect this.

src/lava/magma/core/run_configs.py
srrisbud: The first check is `AbstractSubProcessModel`, the second check is `AbstractPyProcessModel`.. nothing is duplicated there. There can be sub-classes of `AbstractNxProcessModel` in future. So just `Sub` and `Py` do not cover all possible cases.

If `pm` is `Sub` *and* user asked for `Sub`, then we return it in the first `if`. If `pm` is `Py`, we enter the tag-checking if-elif.

src/lava/magma/core/run_configs.py
srrisbud: I have rearranged the class docstring

src/lava/magma/core/run_configs.py
srrisbud: In this case, the pm is not found all the way to the end of the loop. We raise an error. It is a fairly verbose error that a user would be able to know exactly what went wrong.

src/lava/magma/core/run_configs.py
srrisbud: There is only one assertion going wrong here: if we do not find a ProcessModel: either it was not a `Sub`, or not a `Py`, or was a `Py` but did not match the tag criteria.

src/lava/magma/core/run_configs.py
```diff
@@ -67,42 +67,62 @@ def select(self,
...
+            # Priority given to SubProcessModels
+            if issubclass(pm, AbstractSubProcessModel) and \
+                    self.select_sub_proc_model:
+                return pm
+            elif issubclass(pm, AbstractPyProcessModel):
+                # If ProcessModel has tags AND user asked for a specific tag
+                if len(pm.tags) > 0 and (self.select_tag in pm.tags):
+                    return pm
+                # If ProcessModel has no tags and no one asked for a tag anyway
+                elif len(pm.tags) == 0 and not self.select_tag:
```
mathisrichter, line 120: If the user does not ask for a tag (which is the default in Loihi1SimCfg as select_tag=None), and the ProcessModel has at least one tag (like Dense), this currently gives the error:
```
AssertionError: No ProcessModel could be selected with tag 'None' for Process Process_17::Dense.
```
I would assume that some default ProcessModel is chosen if I don't care.

mathisrichter (changes requested):
Found a possible problem while using this feature.

src/lava/magma/core/run_configs.py
awintel: I agree. If there is no ambiguity (not multiple ProcModels with same type for same Proc) then even if the ProcModel has tags but the user does not care about the tag, then we can just pick that one available option.

Ideally we could also detect when there is ambiguity (multiple ProcModels of same type for same Proc) and throw at least a warning when we pick the first ProcModel of many possible matching ones without a tag.

src/lava/magma/core/run_configs.py
```diff
@@ -70,19 +70,21 @@ def select(self,
 class Loihi1SimCfg(RunConfig):
     """Run configuration selects appropriate ProcessModel -- either
     `SubProcessModel` for a hierarchical Process or else a `PyProcessModel`
-    for a standard Process.
+    for a standard Process. Implements and overrides `select()` method of the
```
awintel, line 73: 'Implements AND overrides' is redundant. The parent class is abstract and only provides a virtual select method (we could make it an @abstractmethod). Thus we only implement it.

awintel: It is actually entirely unnecessary to say this because it is a MUST that a concrete child class implements the abstract method of its parent.

src/lava/magma/core/run_configs.py
```diff
@@ -70,19 +70,21 @@ def select(self,
...
+    parent class `RunConfig` to achieve this.
+
+    First, exceptions are checked in `exception_proc_model_map` dictionary
```
awintel, line 76: First explain that this RunCfg has a default behavior of first returning SubProcModel and then LeafProcModels. However this default behavior can be overridden for specific Process classes...

src/lava/magma/core/run_configs.py
```diff
@@ -70,19 +70,21 @@ def select(self,
...
     In case of `PyProcessModel`s, an appropriate `PyProcessModel` is selected
     based on the tags associated with it. Tags are set using the `@tag`
     decorator. If no tags are set for `ProcessModel`s and no tag is selected,
     then the first matching `ProcessModel` will be returned.
-
-    If any exceptions are needed, they can be specified in
-    `exception_proc_model_map` dictionary as `{Process: ProcessModel}`
-    key-value pairs. These explicit specifications are given
-    precedence over the subsequent logic for `ProcessModel` selection"""
+    """
```
awintel, line 87: Also document constructor of this class.

src/lava/magma/core/run_configs.py
```diff
@@ -67,42 +67,60 @@ def select(self,
...
+                 exception_proc_model_map=None):
         super().__init__(custom_sync_domains=custom_sync_domains)
         self.select_tag = select_tag
         self.select_sub_proc_model = select_sub_proc_model
+        self.exception_proc_model_map = exception_proc_model_map
+        if not exception_proc_model_map:
+            self.exception_proc_model_map = {}
 
     def select(self, proc, proc_models):
         from lava.magma.core.model.sub.model import AbstractSubProcessModel
```
awintel, line 102: Add a proper docstring with input argument documentation and type annotations.

src/lava/magma/core/run_configs.py
```diff
@@ -67,42 +67,60 @@ def select(self,
...
-                if self.select_tag in py_proc_model.tags:
-                    return py_proc_model
-            else:
-                raise AssertionError("No legal ProcessModel found.")
+            # Priority given to SubProcessModels
+            if issubclass(pm, AbstractSubProcessModel) and \
+                    self.select_sub_proc_model:
+                return pm
+            elif issubclass(pm, AbstractPyProcessModel):
+                # If ProcessModel has tags AND user asked for a specific tag
```
awintel, line 116: I would put the inline comment describing what a branch does inside the if or elif.

src/lava/magma/core/run_configs.py
```diff
@@ -67,42 +67,60 @@ def select(self,
...
+    based on the tags associated with it. Tags are set using the `@tag`
+    decorator. If no tags are set for `ProcessModel`s and no tag is selected,
+    then the first matching `ProcessModel` will be returned.
+
+    If any exceptions are needed, they can be specified in
+    `exception_proc_model_map` dictionary as `{Process: ProcessModel}`
+    key-value pairs. These explicit specifications are given
+    precedence over the subsequent logic for `ProcessModel` selection"""
+
+    def __init__(self,
```
awintel, line 89: Add type annotations.

awintel (changes requested):
Looks good. Just a few minor requests on documentation and type annotations.

src/lava/magma/core/run_configs.py
```diff
@@ -67,42 +67,60 @@ def select(self,
...
         from lava.magma.core.model.py.model import AbstractPyProcessModel
-        py_proc_model = None
-        sub_proc_model = None
+        # First priority to the exceptions. We will simply return the
+        # ProcessModel class associated with a Process class in the
+        # exceptions dictionary
+        if proc.__class__ in self.exception_proc_model_map:
+            return self.exception_proc_model_map[proc.__class__]
+        # Now, we loop over all ProcessModels proc_models of a Process proc
         for pm in proc_models:
```
awintel, line 110: Actually we made a mistake here:
- What we want is that if there's a SubProcModel, then we want to return it no matter what!
- But currently we are dependent on the order of the ProcModels provided!
- Even if we want the SubProcModel but the PyProcModel comes first, then we will return that!

Instead, we must not return immediately when we iterate through all the proc_models. We must first just store 'pm' as 'selected_pm' during the loop which starts out as None. If we first find a PyProcModel and later encounter a SubProcModel we may still overwrite 'selected_pm'.
Only when 'selected_pm' is still None in the end, do we raise the current AssertionError!

awintel (changes requested):
We made a mistake in the select() method. Check what I wrote near 
'for pm in proc_models'

elvinhajizada (approved):

src/lava/magma/core/run_configs.py
srrisbud: Removed.

src/lava/magma/core/run_configs.py
srrisbud: Done.

src/lava/magma/core/run_configs.py
srrisbud: Done.

src/lava/magma/core/run_configs.py
srrisbud: This is done.

src/lava/magma/core/run_configs.py
srrisbud: Done.

src/lava/magma/core/run_configs.py
srrisbud: Refactored.

src/lava/magma/core/run_configs.py
srrisbud: Done.

src/lava/magma/core/run_configs.py
srrisbud: Now we are checking if number of process models is 1. If it is, we just return that process model without any further checks. We print a warning when doing this, so that the user is aware.

src/lava/magma/compiler/compiler.py
```diff
@@ -194,7 +194,16 @@ def _select_proc_models(
             models: ty.List[ty.Type[AbstractProcessModel]],
             run_cfg: RunConfig) -> ty.Type[AbstractProcessModel]:
         """Selects a ProcessModel from list of provided models given RunCfg."""
-        return run_cfg.select(proc, models)
+        selected_proc_model = run_cfg.select(proc, models)
+        err_msg = f"RunConfig {run_cfg.__class__.__qualname__}.select() must " \
+            f"return a sub-class of AbstractProcessModel. Got" \
+            f" {type(selected_proc_model)} instead."
+        if not isinstance(selected_proc_model, type):
```
mathisrichter, line 201: What does this do, out of curiosity?

tests/lava/proc/monitor/test_monitors.py
```diff
@@ -60,7 +60,7 @@ def test_proc_compiled_without_error(self):
         monitor = Monitor()
         c = Compiler()
         # Compiling should run without error
-        c.compile(monitor, Loihi1SimCfg())
+        c.compile(monitor, Loihi1SimCfg(select_tag='floating_pt'))
```
mathisrichter, line 63: I still think that this should not be required. It makes no sense to tag the `Monitor` Process Model as `floating_pt`.
It's fine if this is a workaround for the release but if we are indeed planning on redesigning the tag system, please add an issue on that.

mathisrichter (changes requested):
Looks good except for the pesky warnings that pop up everywhere whenever you run something. Can we take care of that before the release to make the tutorials and unit tests cleaner?

src/lava/magma/core/run_configs.py
```diff
@@ -70,57 +70,134 @@ def select(self,
...
     If any `SubProcessModel`s are available, then they will be returned first.
 
     In case of `PyProcessModel`s, an appropriate `PyProcessModel` is selected
-    based on the tags associated with it. Tags are set using the `@tag`
-    decorator. If no tags are set for `ProcessModel`s and no tag is selected,
+    based on the tags associated with it (the tags are set using the `@tag`
+    decorator):
+        (a) If no tags are set for `ProcessModel`s and no tag is selected,
     then the first matching `ProcessModel` will be returned.
+        (b) If there is only one ProcessModel for a Process, it will be
```
awintel, line 82: This makes no sense. Imagine there is just a bit-accurate ProcModel but the user explicitly selects bit-approximate then you get something that totally contradicts your selection.

src/lava/magma/core/run_configs.py
```diff
@@ -70,57 +70,134 @@ def select(self,
...
             self.exception_proc_model_map = {}
 
-    def select(self, proc, proc_models):
+    def select(self,
+               proc: AbstractProcess,
+               proc_models: ty.List[ty.Type[AbstractProcessModel]]) \
+            -> ty.Type[AbstractProcessModel]:
+        """
+        Selects an appropriate ProcessModel from a list of ProcessModels for
+        a Process, based on user requests
```
awintel, line 125: requests.

src/lava/magma/core/run_configs.py
```diff
@@ -70,57 +70,134 @@ def select(self,
...
+        Parameters
+        ----------
+        proc: AbstractProcess, Process for which ProcessModel is selected
```
awintel, line 128: var: type, text? Looks strange because it's tempting to read it as a sentence. 

Proper numpy docstring format is
var: type
    Text

src/lava/magma/core/run_configs.py
```diff
@@ -70,57 +70,134 @@ def select(self,
...
+        proc_models: List[AbstractProcessModel], list of ProcessModels of
```
awintel, line 129: ... especially confusing if you continue in lower case like 
var: type, text...

src/lava/magma/core/run_configs.py
```diff
@@ -70,57 +70,134 @@ def select(self,
...
+            Process
+
+        Returns
+        -------
+        Selected ProcessModel class
+        """
         from lava.magma.core.model.sub.model import AbstractSubProcessModel
         from lava.magma.core.model.py.model import AbstractPyProcessModel
+
+        def issubpm(pm: ty.Type[AbstractProcessModel]) -> bool:
```
awintel, line 139: This should be a _private method.

src/lava/magma/core/run_configs.py
```diff
@@ -70,57 +70,134 @@ def select(self,
...
+            """Checks if input ProcessModel is a SubProcessModel"""
+            return issubclass(pm, AbstractSubProcessModel)
+
+        def ispypm(pm: ty.Type[AbstractProcessModel]) -> bool:
```
awintel, line 143: _private

src/lava/magma/core/run_configs.py
```diff
@@ -70,57 +70,134 @@ def select(self,
...
+                  f"available for Process "
+                  f"{proc.name}::{proc.__class__.__qualname__}.")
+            return proc_models[0]
+
+        # If SubProcessModel is asked for
+        if self.select_sub_proc_model:
+            # Get the indices of SubProcessModels in the list of ProcessModels
+            sub_pm_idxs = [idx for idx, pm in enumerate(proc_models) if
+                           issubpm(pm)]
+            if len(sub_pm_idxs) == 0:
```
awintel: No this won't work in general. select_sub_proc_model can only express a priority/preference not a strict requirement. At one point the compiler will have reached the bottom in its search for a ProcModel. If that Process has more than one LeafProcessModel then the prior "if" won't catch it and then this throws an error because there's no infinite continuation for the sub-Process decomposition.

awintel (changes requested):
The select(..) function does not work in general. See detailed inline comment.

I provided some pseudo-code suggestions in the previous review. Did that not work?

Key points:
- The select_proc_model argument only expresses a preference not a strict obligation to select a SubProcModel because in the end, there won't be one.
- We can't just return a ProcModel if its the only one but does not match the explicit tag that the user request. We can do that if the user does not care about tags (has requested none) . But if the user requests something specific we can't return the opposite silently if there's no other choice. Then we must throw an error.

src/lava/magma/compiler/compiler.py
srrisbud: Checks if `selected_proc_model` is a class and not a primitive data type like `str` or `int`

src/lava/magma/core/run_configs.py
```diff
@@ -70,57 +70,134 @@ def select(self,
...
 
     If any `SubProcessModel`s are available, then they will be returned first.
 
     In case of `PyProcessModel`s, an appropriate `PyProcessModel` is selected
-    based on the tags associated with it. Tags are set using the `@tag`
-    decorator. If no tags are set for `ProcessModel`s and no tag is selected,
+    based on the tags associated with it (the tags are set using the `@tag`
+    decorator):
+        (a) If no tags are set for `ProcessModel`s and no tag is selected,
     then the first matching `ProcessModel` will be returned.
```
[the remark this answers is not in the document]
srrisbud: Fixed

src/lava/magma/core/run_configs.py
srrisbud: Right. Punctuation.

src/lava/magma/core/run_configs.py
mathisrichter: If we change this behavior, we also need to change the `Monitor` Process Model, which is currently tagged as `floating_pt`. One of the lava-dnf tutorials is using the `Monitor` Process but selects the tag `fixed_pt`.

src/lava/magma/core/run_configs.py
```diff
@@ -67,42 +67,135 @@ def select(self,
...
+
+        # First priority to the exceptions. We will simply return the
+        # ProcessModel class associated with a Process class in the
+        # exceptions dictionary
+        if proc.__class__ in self.exception_proc_model_map:
+            return self.exception_proc_model_map[proc.__class__]
+
+        # If there is only one ProcessModel available, then we just return
+        # it by default after confirming that it is Sub or Py ProcessModel
+        if len(proc_models) == 1 and (issubpm(proc_models[0]) or ispypm(
```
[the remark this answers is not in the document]
mgkwill: What about something like this:

```
diff --git a/src/lava/magma/core/run_configs.py b/src/lava/magma/core/run_configs.py
index e837c0d..5a86ba9 100644
--- a/src/lava/magma/core/run_configs.py
+++ b/src/lava/magma/core/run_configs.py
@@ -108,13 +108,15 @@ class Loihi1SimCfg(RunConfig):
                  select_sub_proc_model: ty.Optional[bool] = False,
                  exception_proc_model_map: ty.Optional[ty.Dict[
                      ty.Type[AbstractProcess], ty.Type[
-                         AbstractProcessModel]]] = None):
+                         AbstractProcessModel]]] = None,
+                 verbose: ty.Type[bool] = False):
         super().__init__(custom_sync_domains=custom_sync_domains)
         self.select_tag = select_tag
         self.select_sub_proc_model = select_sub_proc_model
         self.exception_proc_model_map = exception_proc_model_map
         if not exception_proc_model_map:
             self.exception_proc_model_map = {}
+        self._verbose = verbose

     def select(self,
                proc: AbstractProcess,
@@ -152,30 +154,33 @@ class Loihi1SimCfg(RunConfig):

         # If there is only one ProcessModel available, then we just return
         # it by default after confirming that it is Sub or Py ProcessModel
-        if len(proc_models) == 1 and (issubpm(proc_models[0]) or ispypm(
-                proc_models[0])):
-            print(f"[{self.__class__.__qualname__}]: Using the only "
-                  f"ProcessModel {proc_models[0].__qualname__} "
-                  f"available for Process "
-                  f"{proc.name}::{proc.__class__.__qualname__}.")
+        if len(proc_models) == 1 \
+            and (_issubpm(proc_models[0])
+                 or _ispypm(proc_models[0])):
+            if self._verbose is True:
+                print(f"[{self.__class__.__qualname__}]: Using the only "
+                      f"ProcessModel {proc_models[0].__qualname__} "
+                      f"available for Process "
+                      f"{proc.name}::{proc.__class__.__qualname__}.")
             return proc_models[0]

         # If SubProcessModel is asked for
         if self.select_sub_proc_model:
             # Get the indices of SubProcessModels in the list of ProcessModels
             sub_pm_idxs = [idx for idx, pm in enumerate(proc_models) if
-                           issubpm(pm)]
+                           _issubpm(pm)]
             if len(sub_pm_idxs) > 0:
                 if len(sub_pm_idxs) > 1:
-                    print(f"[{self.__class__.__qualname__}]: Using the first "
-                          f"SubProcessModel "
-                          f"{proc_models[sub_pm_idxs[0]].__qualname__} "
-                          f"available for Process "
-                          f"{proc.name}::{proc.__class__.__qualname__}.")
+                    if self._verbose is True:
+                        print(f"[{self.__class__.__qualname__}]: Using the "
+                              f"first SubProcessModel "
+                              f"{proc_models[sub_pm_idxs[0]].__qualname__} "
+                              f"available for Process "
+                              f"{proc.name}::{proc.__class__.__qualname__}.")
                 return proc_models[sub_pm_idxs[0]]
```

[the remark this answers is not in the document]
srrisbud: Took care of this. Completely re-wrote the thing again.

src/lava/magma/core/run_configs.py
srrisbud: This is dealt with after second (third?) refactoring.

src/lava/magma/core/run_configs.py
srrisbud: Fixed.

src/lava/magma/core/run_configs.py
srrisbud: Fixed

src/lava/magma/core/run_configs.py
srrisbud: Fixed.

src/lava/magma/core/run_configs.py
srrisbud: _Fixed.

src/lava/magma/core/run_configs.py
srrisbud: This is a review of stale code for some reason. Already caught this and changed it.

tests/lava/proc/monitor/test_monitors.py
srrisbud: Will leave it for now. Let's change the Monitor later.

PR #99, opened 2021-11-23
Tutorial for shared memory access (RefPorts)

PhilippPlank:
Added a tutorial explaining the usage of RefPorts and VarPorts.


awintel (changes requested):
Looks good. Just some minor comments. If you address these, you can just get the sign offs from people in your time zone tomorrow.

- "on a Loihi Neurocore from" -> NeuroCore
- "The read/write of RefPorts are still implemented via messages and can be used across compute resources." Check my tutorial00. Be sure to elaborate a bit more on what this means. You have to say that we do not really access memory of other processes directly . We just expose it as such through the read/write API for convenience. We don't break with the CSP paradigm, thus even read/write is implemented via channels and message passing and the remote process modifies its memory itself based on instructions from another process.
- You could make your prints a little more interesting by printing out the current time steps in front of the value since this is anyways purely for demonstration purposes. Then it's clear to associate the prints with you explanation in the comment.
- Incorrect: "An implicit VarPort is created when connect_var(..) is used to connect a RefPort with a VarPort" An ImplicitVarPort is created when connected to a Var directly.
- The section on "Options to connect RefPorts and VarPorts" Could benefit from a figure showing a hierarchical process. Otherwise it's a bit hard to follow how and why when such cases would arise.
- Any reason for the big space between PyProcModel1 and 2?
- You could also simply demonstrate that RefPorts work the same way hierarchically by constructing a simple hierarchical process.
- In order to make this a deeper dive tutorial than what I'm covering in tutorial00, you could also showcase how RefPort/VarPort interaction relates to SET/GET commands from the Runtime level: You should show that RefPort access takes precedence by doing something with a Var both via SET/GET and RefPort to make the order explicit.

By the way, do RefPorts already support multiplicity like 1:many or many:1? If so? You could give an example. If not, be sure to raise exceptions in the Port objects if a user tries to do that if you have not done that already.

PhilippPlank:
I addressed most of your remarks and added error messages if one tries to make a 1:many/many:1 connection. This is currently not supported and not a trivial change.

Also, it seems that currently using RefPorts with hierarchical processes is not working, it does not read the correct value (have not tried writing yet).

awintel:
So why is that? There should not be any difference if hierarchical processes are involved since the hierarchy gets flattened out anyways. The first question would be does get_dst_ports() perhaps not work correctly across hierarchies?

PhilippPlank:
Fixed the issue, RefPorts and VarPorts work now also with hierarchical processes.

src/lava/magma/compiler/compiler.py
```diff
@@ -206,6 +234,8 @@ def _expand_sub_proc_model(self,
 
         # Build SubProcessModel
         model = model_cls(proc)
+        # Check if VarPorts are configured and propagate them to the sub process
+        self._propagate_var_ports(proc)
```
awintel, line 238: I would probably first check if there are VarPorts and only then call the function to propagate

src/lava/magma/compiler/compiler.py
```diff
@@ -196,6 +196,34 @@ def _select_proc_models(
...
+         directly targeted from processes of a different parent process."""
+
+        for vp in proc.var_ports:
+            v = vp.var.aliased_var
+            if v is not None:
+                sub_proc = v.process
+                # Create an implicit Var port in the sub process
+                new_vp = ImplicitVarPort(v)
+                # Propagate name and parent process of Var to VarPort
+                new_vp.name = "_" + v.name + "_implicit_port"
```
awintel, line 215: I think we did this already somewhere else. Could be factored out into a shared function.

awintel (approved):
Looks fine.
I just have a minor comment about style and a refactoring suggestion. But you can do this later. Does not have functional implications.
Don't want to hold this up any further.

src/lava/magma/compiler/compiler.py
PhilippPlank: correct

mgkwill (approved):

PR #105, opened 2021-11-26
Connection tutorial finished

PhilippPlank:
Modified connection tutorial (tutorial05_connect_processes.ipynb) for releasing.


mathisrichter (approved):
Looks good to me

awintel (approved):
Looks fine for the release. Perhaps just fix a few typos here and there.

Generally, you should come back and improve the narrative of the tutorial a bit. It's not so valuable to just say in English what the code already does. Use English to explain the intent of why you do what you do or explain relevant architectural aspects along the way.

Right now this tutorial does not much deeper than what I will be talking about in the 'tour through lava' tutorial. Once the different PyPorts and VirtualPorts are working, we should actually go deeper on them in this tutorial.

Issue #100, opened 2021-11-23
Recurrent connectivity hangs on execution


mathisrichter:
When creating a simple recurrent network, for instance a LIF that connects to itself via Dense, execution hangs indefinitely.

```python
import unittest
import numpy as np

from lava.magma.core.run_conditions import RunSteps
from lava.magma.core.run_configs import Loihi1SimCfg
from lava.proc.lif.process import LIF
from lava.proc.dense.process import Dense


class TestRecurrentNetwork(unittest.TestCase):
    def test_running_recurrent_network(self):
        """Tests executing an architecture with a recurrent
        connection."""
        num_steps = 10
        shape = (1,)
    
        bias = np.zeros(shape)
        bias[:] = 5000
        lif = LIF(shape=shape, bias=bias, bias_exp=np.ones(shape))
    
        dense = Dense(weights=np.ones((1, 1)))
    
        lif.s_out.connect(dense.s_in)
        dense.a_out.connect(lif.a_in)
    
        lif.run(condition=RunSteps(num_steps=num_steps),
                run_cfg=Loihi1SimCfg())
        lif.stop()
    
        self.assertEqual(lif.runtime.current_ts, num_steps)

```

mathisrichter:
Is this because both `run_spk()` methods of the `PyProcessModel`s of LIF and Dense do a `recv()` as first operation, which blocks until a message comes in - thus blocking the entire network?

From `PyDenseModel`:
```python
    def run_spk(self):
        s_in = self.s_in.recv()
        a_out = self.weights[:, s_in].sum(axis=1)
        self.a_out.send(a_out)
        self.a_out.flush()
```

From `PyLifModelFloat`:
```python
    def run_spk(self):
        a_in_data = self.a_in.recv()
        self.u[:] = self.u * (1 - self.du)
        self.u[:] += a_in_data
        bias = self.bias * (2**self.bias_exp)
        self.v[:] = self.v * (1 - self.dv) + self.u + bias
        s_out = self.v >= self.vth
        self.v[s_out] = 0  # Reset voltage to 0
        self.s_out.send(s_out)
```

ashishrao7:
Did you look into this issue where I had a similar problem https://github.com/lava-nc/lava/issues/32?

mathisrichter:
I wasn't aware of this, thanks Ashish. How did you end up solving it? Could that solution be applied here?

ashishrao7:
[...]

PR #112, opened 2021-11-27
Created floating pt and bit accurate Dense ProcModels + unit tests. Fixes issues #100 and #111.

drager-intel:
Networks with recurrent connectivity architectures no longer deadlock. The bit accurate Dense ProcModel should enable accurate network results for the DNF library and the MNIST tutorial.

Fixes issues #111 and #100  

Eventually we will probably rename this Dense Process to Conn and simply have separate ProcessModels for Dense and Sparse. For now (while there is no implementation of Sparse), I have kept the 'Dense' name for the Process to minimize code changes to our existing tutorials and the DNF library.


src/lava/proc/dense/process.py
```diff
@@ -8,7 +8,7 @@
 
 
 class Dense(AbstractProcess):
-    """Dense connections between neurons. Realizes the following abstract
+    """Dense connections between neurons. Realizes the      following abstract
```
awintel, line 11: We probably have to rethink this again. The documentation below would probably be surprising to the reader because it  reveals knowledge about what's to come in terms of implementation. A Process was meant to be HW agnostic.

This is not ideal:
weights = weights * (2 ** w_scale)
w_scale =  8 - num_weight_bits + weight_exp + isMixed()
In NxSDK so far, we have treated 'weights' as the invariant set of parameters with dynamic range from -256..255. num_wgt_bits was just used to scale the weight increment behind the scenes.
I guess for now it's fine. But we should also revisit this.

src/lava/proc/dense/models.py
```diff
@@ -23,22 +23,23 @@ class PyDenseModelFloat(PyLoihiProcessModel):
     fixed point implementation.
     """
     s_in: PyInPort = LavaPyType(PyInPort.VEC_DENSE, bool, precision=1)
-    a_out: PyOutPort = LavaPyType(PyOutPort.VEC_DENSE,np.float)
-    a_buff: np.ndarray = LavaPyType(np.ndarray,np.float)
-    weights: np.ndarray = LavaPyType(np.ndarray,np.float)
-    weight_exp: float = LavaPyType(float,np.float)
-    num_weight_bits: float = LavaPyType(float,np.float)
-    sign_mode: float = LavaPyType(float,np.float)
+    a_out: PyOutPort = LavaPyType(PyOutPort.VEC_DENSE, np.float)
```
awintel, line 26: Doesn't this lead to a deprecation warning that you are supposed to use float instead of np.float?

src/lava/proc/dense/process.py
```diff
@@ -47,22 +47,22 @@ class Dense(AbstractProcess):
     architectures.
```
awintel, line 47: Deadlock avoidance in recurrent architectures is just one aspect. 
-> Circular buffer that stores output activations accumulated in current time step for future time steps.

src/lava/proc/dense/models.py
```diff
@@ -23,22 +23,23 @@ class PyDenseModelFloat(PyLoihiProcessModel):
...
+    sign_mode: float = LavaPyType(float, np.float)
 
     def run_spk(self):
         # The a_out sent on a each timestep is a buffered value from dendritic
         # accumulation at timestep t-1. This prevents deadlocking in
         # networks with recurrent connectivity structures.
         self.a_out.send(self.a_buff)
         s_in = self.s_in.recv()
-        self.a_buff = self.weights[:,s_in].sum(axis=1)
+        self.a_buff = self.weights[:, s_in].sum(axis=1)
```
awintel, line 39: self.a_buff[:] is probably faster because it does not require repeated memory allocation.

src/lava/proc/dense/models.py
```diff
@@ -23,22 +23,23 @@ class PyDenseModelFloat(PyLoihiProcessModel):
...
+    num_weight_bits: float = LavaPyType(float, np.float)
+    sign_mode: float = LavaPyType(float, np.float)
 
     def run_spk(self):
         # The a_out sent on a each timestep is a buffered value from dendritic
         # accumulation at timestep t-1. This prevents deadlocking in
         # networks with recurrent connectivity structures.
         self.a_out.send(self.a_buff)
         s_in = self.s_in.recv()
-        self.a_buff = self.weights[:,s_in].sum(axis=1)
```
awintel: Leave comment about recurrence away. Even in a feedforward architecture would you want to decouple output generation from waiting till any input is received. If you had a feedforward chain of N processes, then all those had to be processed in sequentially. 
The way we do it allows basically all processes to proceed immediately when a time steps starts.

src/lava/proc/dense/models.py
```diff
@@ -49,42 +50,40 @@ class PyDenseModelBitAcc(PyLoihiProcessModel):
...
-    num_weight_bits: np.ndarray = LavaPyType(np.ndarray,np.int32,precision=3)
-    sign_mode: np.ndarray = LavaPyType(np.ndarray,np.int32,precision=2)
+    weight_exp: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=4)
+    num_weight_bits: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=3)
+    sign_mode: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=2)
 
     def __init__(self):
-        super(PyDenseModelBitAcc,self).__init__()
-        #Flag to determine whether weights have already been scaled.
+        super(PyDenseModelBitAcc, self).__init__()
```
awintel, line 60: Shorter: super().__init__(self)
There shouldn't be any confusion about which parent __init__ to pick.

src/lava/proc/dense/models.py
```diff
@@ -49,42 +50,40 @@ class PyDenseModelBitAcc(PyLoihiProcessModel):
...
-        #Saturate the weights according to the sign_mode:
-            # 0 : null
-            # 1 : mixed
-            # 2 : excitatory
-            # 3 : inhibitory
-        mixed_idx = np.equal(self.sign_mode,1).astype(np.int32)
-        excitatory_idx = np.equal(self.sign_mode,2).astype(np.int32)
-        inhibitory_idx = np.equal(self.sign_mode,3).astype(np.int32)
+        # Saturate the weights according to the sign_mode:
+        # 0 : null
```
awintel, line 68: This is not needed. Why expose this even in the Process documentation?

src/lava/proc/dense/models.py
```diff
@@ -49,42 +50,40 @@ class PyDenseModelBitAcc(PyLoihiProcessModel):
...
+    sign_mode: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=2)
 
     def __init__(self):
-        super(PyDenseModelBitAcc,self).__init__()
-        #Flag to determine whether weights have already been scaled.
+        super(PyDenseModelBitAcc, self).__init__()
+        # Flag to determine whether weights have already been scaled.
         self.weights_set = False
 
     def _set_wgts(self):
```
awintel, line 64: Since there's no learning. This method can just be called during __init__. Even the check for weigths_set during run_spk is unnecessary.

src/lava/proc/dense/models.py
```diff
@@ -102,6 +101,6 @@ def run_spk(self):
         self.a_out.send(self.a_buff)
         s_in = self.s_in.recv()
         a_accum = self.weights[:, s_in].sum(axis=1)
-        self.a_buff = np.left_shift(a_accum,self.weight_exp) if \
-            self.weight_exp > 0 \
-            else np.right_shift(a_accum,-self.weight_exp)
+        self.a_buff = np.left_shift(a_accum,
```
awintel, line 104: Since this stretches over more than one line anyways, a normal if/else statement seems more readable.

tests/lava/proc/dense/test_process.py
```diff
@@ -9,29 +9,21 @@
 
 class TestConnProcess(unittest.TestCase):
     """Tests for Dense class"""
+
     def test_init(self):
         """Tests instantiation of Dense"""
-        shape=(100,200)
-        weights = np.random.randint(100,size=shape)
+        shape = (100, 200)
+        weights = np.random.randint(100, size=shape)
```
awintel, line 16: Please seed random number generator to guarantee deterministic behavior.

tests/lava/proc/dense/test_models.py
```diff
@@ -21,6 +21,7 @@
 class DenseRunConfig(RunConfig):
     """Run configuration selects appropriate Dense ProcessModel based on tag:
     floating point precision or Loihi bit-accurate fixed point precision"""
+
     def __init__(self, custom_sync_domains=None, select_tag='fixed_pt'):
         super().__init__(custom_sync_domains=custom_sync_domains)
```
awintel, line 26: Seems unnecessary: custom_sync_domains=custom_sync_domains
One is enough.

tests/lava/proc/dense/test_models.py
```diff
@@ -48,6 +49,7 @@ class VecSendandRecvProcess(AbstractProcess):
     when there is a True
```
awintel, line 49: Don't we have shared IO classes for this by now? Thought Sumit created them?

tests/lava/proc/dense/test_models.py
```diff
@@ -59,7 +61,7 @@ def __init__(self, **kwargs):
         self.vec_to_send = Var(shape=shape, init=vec_to_send)
         self.send_at_times = Var(shape=(num_steps,), init=send_at_times)
         self.s_out = OutPort(shape=shape)
-        self.a_in = InPort(shape=shape) #enables recurrence test
+        self.a_in = InPort(shape=shape)  # enables recurrence test
 
 
 class VecRecvProcess(AbstractProcess):
```
awintel, line 67: Same here...

tests/lava/proc/dense/test_models.py
```diff
@@ -106,7 +109,7 @@ def run_spk(self):
 @tag('fixed_pt')
```
awintel, line 109: need -> Need

tests/lava/proc/dense/test_models.py
```diff
@@ -152,28 +155,31 @@ def run_spk(self):
 
 class TestDenseProcessModelsFloat(unittest.TestCase):
     """Tests for floating point ProcessModels of Dense"""
+
     def test_float_pm_buffer(self):
         """
-        Tests floating point Dense ProcessModel connectivity and temporal                dynamics. All input 'neurons' from the VecSendandRcv fire once at time           t=4, and only 1 connection weight in the Dense Process is non-zero. The          non-zero connection should have an activation of 1 at timestep t=5.
+        Tests floating point Dense ProcessModel connectivity and temporal
+                   dynamics. All input 'neurons' from the VecSendandRcv fire
```
awintel, line 162: Is there really this strange indentation or is this a rendering error?

tests/lava/proc/dense/test_models.py
```diff
@@ -152,28 +155,31 @@ def run_spk(self):
...
+                   once at time           t=4, and only 1 connection weight
```
awintel, line 163: Why this giant space in the middle of the line?

tests/lava/proc/dense/test_models.py
```diff
@@ -205,15 +212,13 @@ def test_float_pm_fan_in(self):
         send_at_times = np.repeat(False, (num_steps,))
```
awintel, line 212: fan in -> fan-in

tests/lava/proc/dense/test_models.py
```diff
@@ -279,21 +282,19 @@ def test_float_pm_recurrence(self):
         recurrent connectivity architectures.
```
awintel, line 283: Empty line?

tests/lava/proc/dense/test_models.py
```diff
@@ -152,28 +155,31 @@ def run_spk(self):
 
 class TestDenseProcessModelsFloat(unittest.TestCase):
```
awintel, line 156: TestDenseProcessModelFloat

tests/lava/proc/dense/test_models.py
```diff
@@ -308,6 +309,7 @@ def test_float_pm_recurrence(self):
 class TestDenseProcessModelsFixed(unittest.TestCase):
```
awintel, line 308: TestDenseProcessModelFixed

tests/lava/proc/dense/test_models.py
```diff
@@ -308,6 +309,7 @@ def test_float_pm_recurrence(self):
...
     """Tests for fixed point, ProcessModels of Dense, which are bit-accurate
```
awintel, line 309: Why insert this in-between commas? Why plural? You do just test one class.
-> Tests Loihi 1 compatible fixed-point implementation of ProcessModel for Dense Process.

tests/lava/proc/dense/test_models.py
```diff
@@ -308,6 +309,7 @@ def test_float_pm_recurrence(self):
...
     with Loihi hardware"""
+
     def test_bitacc_pm_fan_out_e(self):
         """
         Tests fixed point Dense ProcessModel dendritic accumulation
```
awintel, line 314: fixed-point

awintel: fan-out

awintel (approved):
Few minor docstring and spelling recommendations. Things like fan-out, fan-in, etc. occur multiple times. Did not comment on it each time.

We need to rethink how we write the Dense Process docstring. There's too much knowledge about the implementation in it. But fine for now. I don't want this to gate the release. But we need to revisit this later.

src/lava/proc/dense/process.py
drager-intel: Sounds good. Yes, this implementation was chosen to mirror the implementation in the prototype of Lava, but we should perhaps not expose so much of this implementation at the level of the Process. I have added a ToDo so we remember to address this in a future issue.

src/lava/proc/dense/models.py
drager-intel: Changed all instances of np.float to float.

src/lava/proc/dense/process.py
drager-intel: Changed the docstring to match the suggested description.

src/lava/proc/dense/models.py
srrisbud: > 

We cannot do things in `__init__` of a ProcessModel. The variables are all `LavaPyType`s when we begin. No real computation can happen with these.

srrisbud (approved):

tests/lava/proc/dense/test_models.py
```diff
@@ -0,0 +1,553 @@
...
+        # and 508 such that a_out % 4 = 0.
+        expected_spk_data = np.zeros((num_steps, shape[0]))
+        expected_spk_data[4, :] = [508, -512, 76]
+        self.assertTrue(np.all(expected_spk_data == spk_data_through_run))
+
+    def test_bitacc_pm_fan_out_weight_precision(self):
+        """
+         Tests fixed-point Dense ProcessModel dendritic accumulation
+         behavior when the fan-out of a projecting neuron is greater than 1
+         , connections are both excitatory and inhibitory (sign_mode = 1),
```
mathisrichter, line 443: Same here

src/lava/proc/dense/process.py
```diff
@@ -8,16 +8,63 @@
...
+
+    # ToDo: (DR) Revisit the implementation of w_scale so that less of this
+    # computation is exposed at the level of the Process.
+
     def __init__(self, **kwargs):
-        # super(AbstractProcess, self).__init__(kwargs)
-        # shape = kwargs.pop("shape")
         super().__init__(**kwargs)
         shape = kwargs.get("shape", (1, 1))
+        weights = kwargs.pop("weights", 0)
```
mathisrichter, line 59: This is missing input validation. What if the user inputs a 4D tensor or a 1D vector - does that work? I would expect an exception to be raised immediately if I input data that is of the correct type but, for instance, wrong shape. Same is probably true for other inputs. Adding these would then require unit testing them - which is also missing at the moment.

This is too much to take on now for the release. Please create an issue for this in Github.

src/lava/proc/dense/models.py
```diff
@@ -16,17 +16,91 @@
...
     s_in: PyInPort = LavaPyType(PyInPort.VEC_DENSE, bool, precision=1)
     a_out: PyOutPort = LavaPyType(PyOutPort.VEC_DENSE, np.int32, precision=16)
-    # previously hidden var
+    a_buff: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=16)
     weights: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=8)
+    weight_exp: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=4)
+    num_weight_bits: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=3)
+    sign_mode: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=2)
+
+    def __init__(self):
```
mathisrichter, line 59: This is missing type annotations (here only for the return type of `None`). But this is a more global issue, which should be addressed. Maybe not for the release... But we are piling up technical debt again.

mathisrichter (dismissed):
Looks very nice overall. Very clean and readable. The DNF tutorials still work using these changes, combined with those from PR #96 (tags). I added a few comments, some of which warrant longer discussion or additional stories. Most of them are not crucial for the release.

But address the deprecation warnings, please. And whatever else you feel like taking on now.

If you feel like adding type annotations for Dense, its ProcModels, and the tests, you can do
```
pip install flake8-annotations
flake8 --ignore=ANN101,W503,E501 src/lava/proc/dense
```
or whatever directory/file you want to run it on to show you all linter errors including type annotation errors. Learned that from @bamsumit  :)

awintel (approved):
Looks good

src/lava/proc/dense/models.py
drager-intel: Yes, good idea -- we can come back and readdress this with all existing ProcModel code.

src/lava/proc/dense/process.py
drager-intel: I've added input validation + unit tests of that validation for the shape of 'shape' and 'weights', to ensure the user knows that Dense is expecting a 2D tensor. I will create an issue to add input validation for all other fields of this Process and any existing Processes in the Lava repo.

src/lava/proc/dense/process.py
```diff
@@ -8,16 +8,63 @@
...
-    """Dense connections between neurons.
-    Realizes the following abstract behavior:
-    a_out = W * s_in
+    """Dense connections between neurons. Realizes the      following abstract
+    behavior: a_out = weights * s_in '
+
+    Parameters
+    ----------
+
+    weights:
```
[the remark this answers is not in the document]
drager-intel: I haven't yet addressed this, as this didn't seem to be the convention for our other existing Processes. But if we think all Proc doc strings should look this way, I can file an issue and we can make sure LIF, Conv, Dense, etc are all uniform.

src/lava/proc/dense/process.py
```diff
@@ -8,16 +8,63 @@
...
+    Connection weight matrix.
+
+    weight_exp:
+    Shared weight exponent used to
+    scale magnitude of weights, if needed. Mostly for fixed point
+    implementations. Unnecessary for floating point implementations.
```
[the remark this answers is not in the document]
drager-intel: @awintel commented on this as well. I've added a ToDo so we remember to discuss whether we should hide more of the implementation-specific variables for the bit accurate Dense at the level of the Dense Process.

tests/lava/proc/dense/test_models.py
```diff
@@ -0,0 +1,553 @@
...
+        rcfg = DenseRunConfig(select_tag='floating_pt')
+        dense.run(condition=rcnd, run_cfg=rcfg)
+        dense.stop()
+
+
+class TestDenseProcessModelFixed(unittest.TestCase):
+    """Tests for fixed-point, ProcessModels of Dense, which are bit-accurate
+    with Loihi hardware"""
+
+    def test_bitacc_pm_fan_out_e(self):
```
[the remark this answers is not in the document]
drager-intel: Changed the test name to include the full word 'excitatory'. 

The `bit_accurate_loihi` tag and `fixed_pt` tag are not exactly the same because every ProcModel that is bit accurate with Loihi has fixed point precision, but not every ProcModel with fixed point precision is bit accurate with Loihi. (Conv is one such example of a ProcModel that is currently fixed point precision but not bit accurate with Loihi.) 

In the case of Dense, either the 'fixed_pt' tag or the 'bit_accurate_loihi' tag will tell the RunConfig to choose the PyDenseModelBitAcc ProcModel as Dense's implementation.

src/lava/proc/dense/process.py
```diff
@@ -8,16 +8,63 @@
...
-    Realizes the following abstract behavior:
-    a_out = W * s_in
+    """Dense connections between neurons. Realizes the      following abstract
+    behavior: a_out = weights * s_in '
+
+    Parameters
+    ----------
+
+    weights:
+    Connection weight matrix.
```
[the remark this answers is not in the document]
drager-intel: Great idea! I've added documentation about the expected 2D form of the weight mat and added input validation for the shape of the weight mat.

mgkwill (approved):

Issue #84, opened 2021-11-21
Incomplete PyPort hierarchy


awintel:
Currently, there is an 'unexpected argument' type warning in PyProcBuilder.build() at:
`port = port_cls(csp_ports, pm, p.shape, lt.d_type)`

This is because PyInPort, PyOutPort, ... have constructors like this:
`def __init__(self, csp_recv_ports: ty.List[CspRecvPort], *args):`
`    self._csp_recv_ports = csp_recv_ports`
`    super().__init__(*args)`

... calling the AbstractPortImplementation constructor with the following signature:
`def __init__(`
`    self,`
`    process_model: "AbstractProcessModel",  # noqa: F821`
`    shape: ty.Tuple[int, ...] = tuple(),`
`    d_type: type = int):`

While this works, it is not good style. The Builder only expects an AbstractPyPort which does not have its own overloaded __init__ method but inherits it from AbstractPortImplementation. Thus the Builder throws the 'unexpected argument' warning.
PyInPort and PyOutPort must have  common parent class with an __init__function signature that expects the csp_port passed to it by the Builder.
Then we can also get rid of just accepting an anonymous argument list like *args which is unnecessary.

On a related note: Why is there an AbstractCspSendPort in the type hierarchy between CspSendPort and AbstractCspPort?
- AbstractCspSendPort does nothing but define an abstract send method
- Are there other CspSendPorts inheriting from AbstractCspSendPort?
Same for CspOutPorts...

PhilippPlank:
Add PR #131 to close this issue.

AbstractCspSendPort and AbstractCspRecvPort are not channel specific, while CspSendPort/CspRecvPort seems to be specifically implemented for pypy channels.

PR #131, opened 2021-11-30
Refactoring py/ports

PhilippPlank:
Refactored the creation of implicit var ports into a common function in the RefPort class, which can be used by the compiler or elsewhere in the future.

Refactored inheritance of PyInPorts and PyOutPorts to a common IO parent class and removed the anonymous argument list in the init function. (Issue #84)


mgkwill (approved):
+1

src/lava/magma/core/model/py/ports.py
```diff
@@ -20,7 +21,22 @@ def csp_ports(self) -> ty.List[AbstractCspPort]:
         pass
 
 
-class PyInPort(AbstractPyPort):
+class AbstractIOPyPort(AbstractPyPort):
+    def __init__(self,
+                 csp_ports: ty.List[AbstractCspPort],
+                 process_model: AbstractProcessModel,
+                 shape: ty.Tuple[int, ...] = tuple(),
```
awintel, line 28: Not sure if default arguments should be provided in AbstractClass. They should be provided and documented in most concrete class. Imagine the concrete class allowing non-kwargs arguments and then users will get errors for not initializing some but not other arguments. How confusing would that be.

src/lava/magma/core/model/py/ports.py
```diff
@@ -20,7 +21,22 @@ def csp_ports(self) -> ty.List[AbstractCspPort]:
         pass
 
 
-class PyInPort(AbstractPyPort):
+class AbstractIOPyPort(AbstractPyPort):
```
awintel, line 24: Naming slightly inconsistent with other PyPorts. Others are already called PyInPort, PyOutPort, PyRefPort. The Type of the port and the fact that it is a port are more closely bound than the the fact that there are all kinds of variants like Py, C, Nc, ... Thus it would be logical to call this one AbstractPyIOPort.

src/lava/magma/core/model/py/ports.py
```diff
@@ -66,14 +73,14 @@ class PyInPortVectorDense(PyInPort):
     def recv(self) -> np.ndarray:
         return ft.reduce(
             lambda acc, csp_port: acc + csp_port.recv(),
```
awintel, line 75: Is this a potential performance bottleneck? Should we call reduce when there is only one port?

Is functools reduce the best thing to use rather than a pure numpy function? Not sure if it uses that under the hood.

src/lava/magma/core/model/py/ports.py
```diff
@@ -20,7 +21,22 @@ def csp_ports(self) -> ty.List[AbstractCspPort]:
         pass
```
awintel, line 23: Most classes and send/recv/write/... methods do not have full docstrings with argument documentation. Please complete so we don't have to think about this anymore.

awintel (changes requested):
Looks good. But please make a few minor changes like documentation and naming.

src/lava/magma/core/model/py/ports.py
PhilippPlank: Fixed

src/lava/magma/core/model/py/ports.py
PhilippPlank: Not sure, but this is not the scope of this PR and was not changed.

src/lava/magma/core/model/py/ports.py
PhilippPlank: Added lots of docstrings.

src/lava/magma/core/model/py/ports.py
awintel: I'm not against the arguments as such. I'm questioning the usefulness of default argument assignments in abstract classes.

In particular, what's the purpose of an empty tuple for 'shape'? That does not seem to be a reasonable value anyways. This should probably be a required argument without default. If there's no shape -> Error.

Perhaps a concrete class has a good way of setting default arguments. But it seems an empty tuple is meaningless in a AbstractClass. If not provided, it will later fail somewhere in the weeks because of an empty tuple and it will not be clear why this error shows up at this point.

We should probably even add an assertion or raise an error if an empty tuple is passed...?

src/lava/magma/core/model/py/ports.py
```diff
@@ -14,31 +14,80 @@
 
 
 class AbstractPyPort(AbstractPortImplementation):
+    """Abstract class for ports implemented in python. A port must have one or
```
awintel, line 17: python -> Python (isn't that a proper name?)

The fact it must have one or more csp_ports is not very useful for a person reading this for the first time. What is even a csp_port?

Say what the purpose of the port is. It provides an interface to send and receive messages sent via channels. PyPorts are the implementation of a Port in Python.

awintel: Then say a few words on why there are multiple csp_ports in the first place.

src/lava/magma/core/model/py/ports.py
```diff
@@ -14,31 +14,80 @@
...
+        -------
+        A list of all csp ports used by the port.
+        """
         return self._csp_ports
 
 
-class PyInPort(AbstractIOPyPort):
+class PyInPort(AbstractPyIOPort):
     """Python implementation of InPort used within AbstractPyProcessModel.
     If buffer is empty, recv() will be blocking.
```
awintel, line 76: What buffer?

awintel: Is this actually true? I thought we made it non-blocking in the absence of a DONE token.

By the way, we need a DONE token to be implemented in our CSP channel infrastructure.

src/lava/magma/core/model/py/ports.py
```diff
@@ -14,31 +14,80 @@
...
+
+    Attributes
+    ----------
+    VEC_DENSE : PyInPortVectorDense, default=None
```
awintel, line 80: Not really helpful explanation. This class attribute does not "specify" anything. It's just a class attribute that returns the type of one of four different sub types of PyInputPorts like PyINPortVectorDense, etc.

Then explain what this is even good for: Used to select type of InPorts via LavaPyType declarations in PyProcModels.

awintel: The background explanation should probably pulled out into the main docstring such that in the the Attributes section you only explain what VectorDense, VectorSparse, ...

Also, be sure to say "Class attributes" so it's not confused with instance attributes that are usually defined in __init__ or as part of dataclasses.

src/lava/magma/core/model/py/ports.py
```diff
@@ -115,8 +196,23 @@ def peek(self) -> ty.Tuple[int, int]:
 PyInPort.SCALAR_SPARSE = PyInPortScalarSparse
 
 
-class PyOutPort(AbstractIOPyPort):
-    """Python implementation of OutPort used within AbstractPyProcessModels."""
+class PyOutPort(AbstractPyIOPort):
+    """Python implementation of OutPort used within AbstractPyProcessModels.
+
+    Attributes
```
awintel, line 202: Same here...

src/lava/magma/core/model/py/ports.py
```diff
@@ -160,12 +280,51 @@ def send(self, data: int, idx: int):
...
 
 class VarPortCmd:
+    """Helper class to specify constants."""
     GET = enum_to_np(0)
     SET = enum_to_np(1)
 
 
 class PyRefPort(AbstractPyPort):
-    """Python implementation of RefPort used within AbstractPyProcessModels."""
+    """Python implementation of RefPort used within AbstractPyProcessModels.
```
awintel, line 289: What's a RefPort?

src/lava/magma/core/model/py/ports.py
```diff
@@ -160,12 +280,51 @@ def send(self, data: int, idx: int):
...
+    process_model : AbstractProcessModel
+        The process model used by the process of the port.
+
+    shape : tuple, default=tuple()
+        The shape of the port.
+
+    d_type: type, default=int
+        The data type of the port.
+
+    Attributes
```
awintel, line 308: Same here... 
+Class attributes

src/lava/magma/core/model/py/ports.py
```diff
@@ -261,6 +461,49 @@ def write(self, data: int, idx: int):
...
```
awintel, line 485: Same here...
+class attributes

src/lava/magma/core/model/py/ports.py
```diff
@@ -261,6 +461,49 @@ def write(self, data: int, idx: int):
 
 class PyVarPort(AbstractPyPort):
     """Python implementation of VarPort used within AbstractPyProcessModel.
```
awintel, line 463: What is a VarPort?

src/lava/magma/core/model/py/ports.py
```diff
@@ -282,22 +525,32 @@ def __init__(self,
...
+        """
         if self._csp_send_port is not None and self._csp_recv_port is not None:
             return [self._csp_send_port, self._csp_recv_port]
         else:
             # In this case the port was not connected
             return []
 
+    @abstractmethod
     def service(self):
+        """Abstract method to set the received value to the given Var or sends
```
awintel, line 542: Incomplete explanation. What received value? For this to make sense, you have to give a refresher what a VarPort is and how it work. Not super long but briefly.

src/lava/magma/core/model/py/ports.py
```diff
@@ -282,22 +525,32 @@ def __init__(self,
...
         pass
 
 
 class PyVarPortVectorDense(PyVarPort):
+    """Python implementation of VarPort for dense vector data."""
     def service(self):
-        """Sets the received value to the given var or sends the value of the
-        var to the csp_send_port, depending on the received header information
-        of the csp_recv_port."""
+        """Abstract method to set the received value to the given Var or sends
```
awintel, line 551: Why still abstract?

src/lava/magma/core/model/py/ports.py
```diff
@@ -348,7 +610,14 @@ def peek(self) -> ty.Tuple[int, int]:
 
 
 class RefVarTypeMapping:
-    """Class to get the mapping of PyRefPort types to PyVarPortTypes."""
+    """Class to get the mapping of PyRefPort types to PyVarPort types.
```
awintel, line 613: Another sentence what this is good for or why it exists?

awintel (changes requested):
Thanks.

I suggest to rework some of the docstrings. We should not just write "something". Put yourself into the shoes of some random person who's reading our code. What needs to be said about a class or method so that it makes sense what a class is, and what it's reason for existence is.

If we don't say that, what good is any other documentation?

src/lava/magma/core/model/py/ports.py
PhilippPlank: Oh okay, I overlooked the default part. Removed the defaults.

src/lava/magma/core/model/py/ports.py
PhilippPlank: recv is still blocking afaik, if there is no data. Each send needs a recv and vice versa in order to not block execution.

I removed this comment in the class description (it was a leftover from before) and added the information in the docstring of recv().

awintel: Sorry. Now I remember, recv() is blocking when connected but when there is not data. However, if the port is not connected then it should not be blocking.

src/lava/magma/core/model/py/ports.py
PhilippPlank: Done

src/lava/magma/core/model/py/ports.py
PhilippPlank: Added explanation of RefPorts, what they do and how they work.

src/lava/magma/core/model/py/ports.py
PhilippPlank: Done

src/lava/magma/core/model/py/ports.py
PhilippPlank: Added explanation of a VarPort, what they do and how they work.

src/lava/magma/core/model/py/ports.py
PhilippPlank: Added explanation.

src/lava/magma/core/model/py/ports.py
PhilippPlank: Fixed.

src/lava/magma/core/model/py/ports.py
```diff
@@ -26,36 +26,40 @@ class AbstractPyPort(AbstractPortImplementation):
     @abstractmethod
     def csp_ports(self) -> ty.List[AbstractCspPort]:
         """
-        Abstract property to get all csp_ports of a PyPort.
+        Abstract property to get all connected PyPorts (csp_ports) of a PyPort.
```
awintel, line 29: Does not seem right. csp_ports are the fundamental CSP ports that implement the CSP protocol. PyPorts are not the csp_ports.

src/lava/magma/core/model/py/ports.py
```diff
@@ -16,8 +16,8 @@
 class AbstractPyPort(AbstractPortImplementation):
     """Abstract class for Ports implemented in Python.
 
-    It provides an interface to send and receive messages sent via channels. The
-    Communicating Sequential Processes (CSP) paradigm is followed in this
+    It provides an interface to send and receive messages via channels. The
+    Communicating Sequential Processes (CSP) paradigm is the inspiration of this
```
awintel, line 20: The difference between CSP Ports and PyPorts should be spelled out.

src/lava/magma/core/model/py/ports.py
```diff
@@ -26,36 +26,40 @@ class AbstractPyPort(AbstractPortImplementation):
...
         pass
 
 
 class AbstractPyIOPort(AbstractPyPort):
-    """Abstract class of an input/output port implemented in python.
+    """Abstract class of an input/output Port implemented in python.
+
+    A PyIOPort can either be an input or an output Port and is the common
+    abstraction of PyInPort/PyOutPort.
+    _csp_ports is a list of the connected PyIOPorts.
```
awintel, line 43: No

src/lava/magma/core/model/py/ports.py
awintel: Perhaps useful to explain the relation between Ports and PyPorts. Ports at the Proc level provide an interface to connect ports. PyPorts provide an implementation for message exchange in Python. That is they provide send, receveive interfaces for OutPorts and InPorts, etc.

src/lava/magma/core/model/py/ports.py
```diff
@@ -208,19 +231,27 @@ def peek(self) -> ty.Tuple[int, int]:
...
 
-    Attributes
-    ----------
+    PyOutPort is an output Port processing data sent from a connected input Port
+    (PyInPort) over a channel. PyOutPort can send (send()) the data, which
+    adds it to the channel or clear (flush()) the channel to remove any data of
+    the channel.
+    The different class attributes are used to select the type of OutPorts via
+    LavaPyType declarations in PyProcModels.
+    A PyOutPort (source) can be connected to one or multiple PyInPorts (target).
```
awintel, line 240: Give an example of a LavaPyType so it becomes clear what you mean?

src/lava/magma/core/model/py/ports.py
```diff
@@ -289,50 +320,61 @@ def send(self, data: int, idx: int):
...
     GET = enum_to_np(0)
     SET = enum_to_np(1)
 
 
 class PyRefPort(AbstractPyPort):
     """Python implementation of RefPort used within AbstractPyProcessModels.
 
+    A PyRefPort is a Port connected to a VarPort of a variable Var of another
+    Process. It is used to get or set the value of the referenced Var across
+    Processes. A PyRefPort is connected via two channels to a PyVarPort. One
```
awintel, line 334: CSP channels

awintel (changes requested):
A few small mistakes in CSP port documentation.

awintel:
@phstratmann I believe you are creating a template for the ProcLib? Would you useful if you looked at the PyPort documentation as well with fresh eyes to make sure we have documented everything that a "newcomer" might wonder about.

src/lava/magma/core/model/py/ports.py
PhilippPlank: Fixed

src/lava/magma/core/model/py/ports.py
PhilippPlank: Fixed

src/lava/magma/core/model/py/ports.py
```diff
@@ -9,96 +9,229 @@
...
+    inspiration of the messaging infrastructure. Thus, a channel denotes a CSP
+    channel of the messaging infrastructure and CSP Ports denote the low level
+    ports also used in the messaging infrastructure. PyPorts are the
+    implementation for message exchange in Python, using the low level CSP Ports
+    of the backend messaging infrastructure. PyPorts provide interface for send
+    and receive of input and output Ports. A PyPort may have one or multiple
+    connection to other PyPorts. These connections are represented by csp_ports,
+    which is a list of CSP ports corresponding to the connected PyPorts.
+    """
     @property
```
phstratmann, line 32: Some suggestions:

Ports at the Process level provide an interface to connect Processes with each other. **Once two Processes have been connected by Ports, they can exchange data. \newline
Lava provides four types of Ports: _briefly describe the four Port types and what they can do_.\newline
The exchanged data can be of either of four types: _briefly explain the difference between VEC_DENSE, VEC_SPARSE , ..._.\newline
To exchange data,** PyPorts provide an interface to send and
    receive messages via channels implemented by a backend messaging
    infrastructure**, which has been inspired by the** Communicating Sequential Processes (CSP) paradigm ~~is the
    inspiration of the messaging infrastructure~~. Thus, a channel denotes a CSP
    channel of the messaging infrastructure and CSP Ports denote the low level
    ports also used in the messaging infrastructure. PyPorts are the
    implementation for message exchange in Python, using the low level CSP Ports
    of the backend messaging infrastructure. ~~PyPorts provide interface for send
    and receive of input and output Ports.~~ A PyPort may have one or multiple
    connection to other PyPorts. These connections are represented by csp_ports,
    which is a list of CSP ports corresponding to the connected PyPorts.

src/lava/magma/core/model/py/ports.py
```diff
@@ -9,96 +9,229 @@
...
+        """
+        Abstract property to get the corresponding CSP Ports of all connected
+        PyPorts (csp_ports). The CSP Port is the low level interface of the
+        backend messaging infrastructure which is used to send and receive data.
+
+        Returns
+        -------
+        A list of all CSP Ports connected to the PyPort.
+        """
         pass
```
phstratmann, line 44: What do you mean by "the corresponding CSP Ports of all connected PyPorts"?
Maybe "Abstract property of a PyPort that returns a list csp_ports of all CSP ports of PyPorts that this PyPort is connected to."?

src/lava/magma/core/model/py/ports.py
```diff
@@ -9,96 +9,229 @@
...
+
+    PyInPort is an input Port processing data sent from a connected output Port
+    (PyOutPort) over a channel. PyInPort can receive (recv()) the data, which
+    removes it from the channel, look at (peek()) at the data which keeps it on
+    the channel or check (probe()) if there is data on the channel.
+    The different class attributes are used to select the type of OutPorts via
+    LavaPyType declarations in PyProcModels, e.g., LavaPyType(
+    PyInPort.VEC_DENSE, np.int32, precision=24) creates a PyInPort.
+    A PyOutPort (source) can be connected to one or multiple PyInPorts (target).
+
```
phstratmann, line 107: May not be fully clear what "processing data" means.
What's about: "PyInPort is an input Port that can be instantiated by a Process to receive data sent from a connected ->PyOutPort of another Process over a channel."

"look ~~at~~ (peek()) at the data"

Unclear why it is import to remove data from a channel, should be explained in a sentence.

"A PyOutPort (source) can be connected to one or multiple PyInPorts (target)."
This docstring is about PyInPorts. Thus, I would here turn it around so that PyInPorts are the important aspects of this sentence:
"Multiple PyInPorts (target) can be connected to the same PyOutPort (source).

src/lava/magma/core/model/py/ports.py
```diff
@@ -9,96 +9,229 @@
...
-        """Returns all csp ports of the port."""
-        return self._csp_recv_ports
-
     @abstractmethod
     def recv(self):
+        """Abstract method to receive data (vectors/scalars) sent from connected
+        out Ports (source Ports). Removes the retrieved data from the channel.
+        Expects data on the channel and will block if there is no data to
+        retrieve on the channel.
+
```
phstratmann, line 134: "will block the process from progressing to the next time step if there is no"...?

src/lava/magma/core/model/py/ports.py
```diff
@@ -9,96 +9,229 @@
...
 
 
 class PyInPortVectorDense(PyInPort):
+    """Python implementation of PyInPort for dense vector data."""
     def recv(self) -> np.ndarray:
+        """Method to receive data (vectors/scalars) sent from connected
+        out Ports (source Ports). Removes the retrieved data from the channel.
+        Expects data on the channel and will block if there is no data to
+        retrieve on the channel.
+
```
phstratmann, line 177: One general comment: Sometimes, I see the term "out Ports" or similar. Sometimes, it is "OutPorts". I think we should be consistent.

src/lava/magma/core/model/py/ports.py
```diff
@@ -108,50 +241,89 @@ def peek(self) -> ty.Tuple[int, int]:
...
+    PyOutPort is an output Port processing data sent from a connected input Port
+    (PyInPort) over a channel. PyOutPort can send (send()) the data, which
+    adds it to the channel or clear (flush()) the channel to remove any data of
+    the channel.
+    The different class attributes are used to select the type of OutPorts via
+    LavaPyType declarations in PyProcModels, e.g., LavaPyType(
+    PyOutPort.VEC_DENSE, np.int32, precision=24) creates a PyOutPort.
+    A PyOutPort (source) can be connected to one or multiple PyInPorts (target).
+
```
phstratmann, line 255: "PyOutPort is an output Port processing data sent from a connected input Port (PyInPort) over a channel."
I thought that an OutPort sends data, and does not receive data from an InPort?

"PyOutPort can send (send()) the data **by adding it to the channel**, or **it can** clear (flush()) the channel to remove any data **from it.**"

src/lava/magma/core/model/py/ports.py
```diff
@@ -108,50 +241,89 @@ def peek(self) -> ty.Tuple[int, int]:
...
 class PyOutPortVectorDense(PyOutPort):
+    """Python implementation of PyOutPort for dense vector data."""
+
     def send(self, data: np.ndarray):
-        """Sends data only if port is not dangling."""
-        for csp_port in self._csp_send_ports:
+        """Abstract method to send data to the connected in Port (target).
+
+        Sends data only if Port is not dangling.
+
```
phstratmann, line 299: For me, "dangling" always has a negative connotation and imply memory safety violations. Here, it shouldn't be that severe. Thus, we may use "only if the OutPort is connected to at least one InPort."?

src/lava/magma/core/model/py/ports.py
```diff
@@ -162,12 +334,62 @@ def send(self, data: int, idx: int):
...
+    """Python implementation of RefPort used within AbstractPyProcessModels.
+
+    A PyRefPort is a Port connected to a VarPort of a variable Var of another
+    Process. It is used to get or set the value of the referenced Var across
+    Processes. A PyRefPort is connected via two CSP channels and corresponding
+    CSP ports to a PyVarPort. One channel is used to send data from the
+    PyRefPort to the PyVarPort and the other channel is used to receive data
+    from the PyVarPort. PyRefPorts can get the value of a referenced Var
+    (read()) or set the value of a referenced Var (write()).
+
```
phstratmann, line 353: Great explanation, thanks! To make it fully obvious, I would add, as second sentence, the precise difference to In/OutPorts.

src/lava/magma/core/model/py/ports.py
```diff
@@ -108,50 +241,89 @@ def peek(self) -> ty.Tuple[int, int]:
...
     def send(self, data: int):
+        """TBD"""
         pass
 
 
 class PyOutPortScalarSparse(PyOutPort):
+    """Python implementation of PyOutPort for sparse scalar data."""
     def send(self, data: int, idx: int):
+        """TBD"""
         pass
```
phstratmann, line 329: Is the code in line 330-333 actually correct? Why do we have these definitions outside any class?

src/lava/magma/core/model/py/ports.py
```diff
@@ -176,27 +398,44 @@ class PyRefPort(AbstractPyPort):
...
     def read(
             self,
     ) -> ty.Union[
         np.ndarray, ty.Tuple[np.ndarray, np.ndarray], int, ty.Tuple[int, int]
     ]:
+        """Abstract method to request and return data from a VarPort.
+        Returns
+        -------
+        The value of the referenced var.
+        """
```
phstratmann, line 435: I would always add the data type of the return var.

src/lava/magma/core/model/py/ports.py
```diff
@@ -260,6 +530,59 @@ def write(self, data: int, idx: int):
...
 class PyVarPort(AbstractPyPort):
     """Python implementation of VarPort used within AbstractPyProcessModel.
+
+    A PyVarPort is a Port related to a variable Var of a Process and might be
+    connected to a RefPort of another process. It is used to get or set the
+    value of the referenced Var across Processes. A PyVarPort is connected via
+    two channels to a PyRefPort. One channel is used to send data from the
+    PyRefPort to the PyVarPort and the other is used to receive data from the
+    PyVarPort. PyVarPorts set or send the value of the related Var (service())
+    given the command VarPortCmd received by a connected PyRefPort.
```
phstratmann, line 540: Here and below, I would replace "related var" -> "linked var".

src/lava/magma/core/model/py/ports.py
```diff
@@ -270,30 +593,49 @@ class PyVarPort(AbstractPyPort):
...
+            # In this case the Port was not connected
             return []
 
+    @abstractmethod
     def service(self):
+        """Abstract method to set the value of the related Var of the VarPort,
+        received from the connected RefPort, or to send the value of the related
+        Var of the VarPort to the connected RefPort. The connected RefPort sends
+        a command VarPortCmd first to determine a read() or write() operation.
+        """
```
phstratmann, line 627: The connected RefPort determines whether it will perform a read() or write() operation by sending a command VarPortCmd.

src/lava/magma/core/model/py/ports.py
```diff
@@ -344,7 +707,20 @@ def peek(self) -> ty.Tuple[int, int]:
 
 
 class RefVarTypeMapping:
-    """Class to get the mapping of PyRefPort types to PyVarPortTypes."""
+    """Class to get the mapping of PyRefPort types to PyVarPort types.
```
phstratmann, line 710: Is the code in line 703 actually correct? Why do we need to specify these values at that point outside any class?

src/lava/magma/core/process/ports/ports.py
```diff
@@ -46,11 +46,11 @@ def __init__(self, shape: ty.Tuple):
         self.out_connections: ty.List[AbstractPort] = []
```
phstratmann, line 46: I very much like the docstring of AbstractPort! However, I would add a few information:
Maybe a first sentence like "Ports are the main building block of Lava to exchange data between different processes."

For the sentence "Sub classes of AbstractPort only facilitate connecting to other compatible ports"
I would state the actual connections that are possible (OutPort -> Inport; VarPort -> RefPort).

"a port may have zero or more input and output connections"
This sounds as if every Port can have both input and output connections. But an InPort cannot have one or multiple Input connections.

"These connections, imply an a-cyclic graph structure that allows the compiler to infer connections between processes."
This sounds as if we cannot build cyclig graphs (like RNN) in Lava. Not sure what is exactly meant with this statement?

src/lava/magma/core/process/ports/ports.py
```diff
@@ -181,9 +181,9 @@ def flatten(self) -> "ReshapePort":
         return self.reshape((self.size,))
```
phstratmann, line 181: Regarding line 145-146:
That sounds as if you actually cycle through the whole connectivity graph. I.e., if port a is connected to port b of Process B. and Process B via ports b2->c, then I would expect port c to be returned as well according to the docstring. But I guess that's not what is meant here?

src/lava/magma/core/process/ports/ports.py
```diff
@@ -254,7 +254,7 @@ class OutPort(AbstractIOPort, AbstractSrcPort):
     """
```
phstratmann, line 254: Regarding AbstractSrcPort:
Maybe a stupid question, but I thought RefPorts are bi-directional. Thus, why can we label them as Source Port?

src/lava/magma/core/process/ports/ports.py
```diff
@@ -339,7 +339,7 @@ class RefPort(AbstractRVPort, AbstractSrcPort):
     RefPort to a Var via the connect_var(..) method."""
```
phstratmann, line 339: "Shared-memory-based communication can have side-effects and should therefore be used with caution."

Would it be possible to elaborate on this? As a user, I will be alerted after reading this sentence, but will not know what to do with it.

src/lava/magma/core/model/py/ports.py
PhilippPlank: Thanks for the suggestions, I used some of it. I will not explain VEC_DENSE etc. here, because these types are independent of the abstract class and vice versa -> they can change without adapting this class at all.
Same thing for shape.

src/lava/magma/core/model/py/ports.py
PhilippPlank: Only the corresponding CSP Ports used for the connection not all CSP Ports. Changed it to: "Abstract property to get a list of the corresponding CSP Ports of all connected PyPorts."

src/lava/magma/core/model/py/ports.py
```diff
@@ -9,96 +9,229 @@
...
+    A PyIOPort can either be an input or an output Port and is the common
+    abstraction of PyInPort/PyOutPort.
+    _csp_ports is a list of CSP Ports which are used to send/receive data by
+    connected PyIOPorts.
+
+    Parameters
+    ----------
+    csp_ports : list
+        A list of csp Ports used by this IO Port.
+
```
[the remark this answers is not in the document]
PhilippPlank: Changed.

Yes, csp_ports is a parameter, which the user had to provide on call and _csp_ports is an attribute of the class. They hold the same information, but are not the same thing.

src/lava/magma/core/model/py/ports.py
PhilippPlank: Applied some suggestions

src/lava/magma/core/model/py/ports.py
PhilippPlank: Changed to "will block execution if there is no data to retrieve on the channel"

src/lava/magma/core/model/py/ports.py
```diff
@@ -9,96 +9,229 @@
...
-        bool values with AND operation. The accumulator acc is initialized to
-        True.
+        """Method to check (probe) if there is data (vectors/scalars)
+        to receive from connected out Ports (source Ports).
+
+        Returns
+        -------
+        result : bool
+             Returns True only when probe returns True for all connected OutPort
+             channels.
```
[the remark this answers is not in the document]
PhilippPlank: Done

src/lava/magma/core/model/py/ports.py
PhilippPlank: Changed all "out p/Ports" to "OutPorts"

src/lava/magma/core/model/py/ports.py
PhilippPlank: Changed

src/lava/magma/core/model/py/ports.py
PhilippPlank: The datatype is specified by the type hint, so it should not be necessary.

src/lava/magma/core/model/py/ports.py
PhilippPlank: Done

src/lava/magma/core/model/py/ports.py
PhilippPlank: Changed

src/lava/magma/core/model/py/ports.py
PhilippPlank: see answer above

src/lava/magma/core/process/ports/ports.py
PhilippPlank: This is the most abstract description and definition of a Port we have in Lava. In/Out or Ref/Var Ports are much more specialized and should not be mentioned here at all.
Maybe someone adds different Port types in the future.

Also, this docstring is from Andreas I think, I have not touched docstrings in this file and have little knowledge of what is going on there. So I prefer to not alter the docstring in this file in this PR.

src/lava/magma/core/process/ports/ports.py
PhilippPlank: see answer above

awintel (approved):

src/lava/magma/core/process/ports/ports.py
PhilippPlank: see above

src/lava/magma/core/model/py/ports.py
phstratmann: Looks great, thanks!

phstratmann (changes requested):
One minor comment concerning VECTOR_DENSE, VECTOR_SPARSE,...
Then I'll be happy to approve.

Thanks for your work!

src/lava/magma/core/model/py/ports.py
```diff
@@ -9,96 +9,244 @@
...
+
+    VEC_SPARSE : PyInPortVectorSparse, default=None
+        Type of PyInPort. Used for sparse vector data.
+
+    SCALAR_DENSE : PyInPortScalarDense, default=None
+        Type of PyInPort. Used for dense scalar data.
+
+    SCALAR_SPARSE : PyInPortScalarSparse, default=None
+        Type of PyInPort. Used for sparse scalar data.
     """
```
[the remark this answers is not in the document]
PhilippPlank: Good question, I do not know, I have not written these placeholders. Only VEC_DENSE is currently implemented and usable.

@awintel What should SCALAR_SPARSE be?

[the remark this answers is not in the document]
PhilippPlank: Added explanation to the docstrings

phstratmann (approved):
Thanks for your great revision!

Issue #83, opened 2021-11-21
Runtime deadlocks when ProcessModel building or execution fails for any reason without any exceptions thrown at user


awintel:
None of our unit tests so far has detected, that any errors happening during building of a remote ProcessModel or during execution do not get thrown to the user process. 

Instead, the Runtime just deadlocks and it is very hard to figure out where the problem happened. 

Presumably, this is because the remote system process, running the ProcessModel, dies and thus the Runtime or RuntimeService waits forever on a channel response that the ProcessModel is done doing what it was tasked to do.

Perhaps, if we detect that something failed in the remote system process and communicate this back via an ERROR management token, then the system can terminate gracefully and may even print any exceptions that get thrown.

Aside from fixing this issue, we should also create a ProcessTester utility which wraps around any individual ProcessModel (without a Compiler and Runtime), creates complementary Ports to any ProcessModel Ports and allows to inject targeted messages into the process in order to debug it without having to set up the entire machinery of Compiler and Runtime.

PR #135, opened 2021-11-30
Runtime error handling

PhilippPlank:
Solving the issue of deadlock in the runtime if any ProcessModel throws an exception. (Issue #83)
Also passing the exception all the way up to the runtime. (model -> runtime_service -> runtime)

Open for discussion is the usage of existing csp_ports for error messages or if there are better ways to transport the exception to the runtime.

Edit: the unit test currently does not really work as unit test, but shows a use case.


PhilippPlank:
Updated initial commit.

I come across the channel size which is 64 at the moment for pypy channels. Why is it 64? I needed to increase it a bit in order to send a longer error message over the channels. 
There are other options of course (starting the read at the same time etc.), but I wanted to keep it simple for now.

I think this should work now and stop all running process model threads, if there are one or multiple errors in the process models.

awintel:
Wouldn't a simply pipe like in this example be simpler through which you can just push the exception as is?
https://newbedev.com/python-multiprocessing-handling-child-errors-in-parent
If you rethrow it at the Runtime level, it will probably print the usual nice stack trace so you can click on the hyperlinks in your IDE which will take you to the place in the code that caused the problem.

PhilippPlank:
I adopted the behavior to your suggestion @awintel 
This also allows unit tests to actually succeed. 
I am quite happy with the new solution, let me know if there is anything else.

joyeshmishra (approved):

mgkwill (approved):
+1

src/lava/magma/runtime/runtime.py
```diff
@@ -229,25 +229,19 @@ def _run(self, run_condition):
...
-
-                                raise Exception(exceptions)
+                                # Receive all errors from the ProcessModels
+                                error_cnt = 0
+                                for actors in \
+                                        self._messaging_infrastructure.actors:
+                                    actors.join()
+                                    if actors.exception:
+                                        _, traceback = actors.exception
+                                        print(traceback)
```
awintel, line 239: Does this actually identify the problem from which the error came? Perhaps attach this to the Exception object.
You should print the class name, Process name and id.

awintel: To be clear. Both the processes should identify themselves but also the RuntimeServices in case the error happened in the RuntimeService.
Perhaps you even want to distinguish them clearly in the command line prints.

Perhaps you even want to first print a summary of everything that has thrown any errors at the top of the command line to orient the user because otherwise there could be quite a messy stack trace.

src/lava/magma/runtime/runtime.py
```diff
@@ -229,25 +229,19 @@ def _run(self, run_condition):
...
+                                        error_cnt += 1
+
+                                raise RuntimeError(
```
awintel, line 242: You could define own Execptions. RuntimeError for actual errors in the Runtime and ProcessModelError for ... you know what.

awintel (commented):
Looks good. I would just suggest to give a bit more detailed feedback to the user.

src/lava/magma/runtime/runtime.py
PhilippPlank: The printed traceback leads you to the line of code the exception occurred, e.g., "line 50, in run_spk
    raise AssertionError("All the error info")" -> line 50 is within PyProcModel1. So you know exactly which exception happened and on which code line.

These tracebacks are printed for every exception occurred in all ProcessModels.

src/lava/magma/runtime/runtime.py
PhilippPlank: This exception is a RuntimeError, due to other Exceptions happened in the ProcessModel. The details have been printed already, this is just to stop the Runtime and tell the user that there have been other Exceptions (+ the number of them). 

For all the details the user only has to scroll upwards in the console.

PR #180, opened 2022-02-15
Sigma delta neuron, encoding and decoding support

bamsumit:
Issue Number: 

Objective of pull request:

## Pull request checklist

## Pull request type

 

 

Please check your PR type:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] Documentation changes

- There is no support for sigma delta neuron.

## What is the new behavior?

- Sigma Delta neuron with Unit and ReLU activation is supported for both floating and fixed point.

- [ ] Yes
- [x] No


src/lava/proc/sdn/process.py
```diff
@@ -0,0 +1,150 @@
...
+from enum import IntEnum, unique
+
+from lava.magma.core.process.process import AbstractProcess
+from lava.magma.core.process.variable import Var
+from lava.magma.core.process.ports.ports import InPort, OutPort
+
+
+@unique
+class ACTIVATION_MODE(IntEnum):
+    """Enum for synapse sigma delta activation mode. Options are {``ReLU : 0``}.
```
srrisbud, line 15: 1. ReLU is 1. So the dict in the docstring should say {``ReLU : 1``}
2. Are we planning to extend this to more activations? If yes, maintaining an Enum will get unwieldy beyond 4 or 5 neuron models.
3. Also, "ACTIVATION_MODE" sort of makes SigmaDelta as the "neuron model" with just 2 activation modes. If we want to apply Sigma and Delta around an arbitrary neuron model, we can all this Enum something else.

tests/lava/proc/sdn/test_models.py
```diff
@@ -0,0 +1,265 @@
...
+            state_exp=state_exp,
+            cum_error=False,
+            tag='floating_pt'
+        )
+
+        error = np.abs(input - output).max()
+
+        if verbose:
+            print(f'Max abs error = {error}')
+    #     self.assertTrue(error < vth * (1 << wgt_exp))
```
srrisbud, line 167: Why is the assertion commented out?

tests/lava/proc/sdn/test_models.py
```diff
@@ -0,0 +1,265 @@
...
+        )
+
+        error = np.abs(np.maximum(input, 0) - output).max()
+
+        if verbose:
+            print(f'Max abs error = {error}')
+        self.assertTrue(error < vth * (1 << wgt_exp))
+
+
+if __name__ == '__main__':
```
srrisbud, line 264: I don't think we are doing `if __name__ == '__main__'` in all unit-tests. I am guessing you are doing this to run tests through command line?

srrisbud (changes requested):

src/lava/proc/sdn/process.py
bamsumit: That's correct. fixed ReLU enum. We have individual Sigma and Delta processes as well which can be wrapped around any arbitrary process. However, for eventual implementation in ucode, we would need separate implementation for each of them. So the thinking is we can use the enum to switch between different ucode implementations that we support and extend the enums accordingly.

tests/lava/proc/sdn/test_models.py
bamsumit: My bad. Fixed it.

tests/lava/proc/sdn/test_models.py
bamsumit: Yeah. I'll remove it.

src/lava/proc/sdn/models.py
```diff
@@ -0,0 +1,296 @@
...
+
+    vth = None
+    sigma = None
+    act = None
+    residue = None
+    error = None
+    bias = None
+    wgt_exp = None
+    state_exp = None
+    cum_error = None
```
mgkwill, line 109: Perhaps this is for clarity but we are redefining these here, as they are also defined in the base classes.

src/lava/proc/sdn/models.py
```diff
@@ -0,0 +1,296 @@
...
+
+    sigma: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=24)
+
+    def __init__(self, proc_params: Dict[str, Any]) -> None:
+        super().__init__(proc_params)
+
+    def run_spk(self) -> None:
+        a_in_data = self.a_in.recv()
+        self.sigma = self.sigma_dynamics(a_in_data)
+        self.s_out.send(self.sigma)
```
mgkwill, line 188: These two classes are the same except for `a_in` and `s_out`. Perhaps a base class that the two inherit from and only change the ports?

src/lava/proc/sdn/models.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+@implements(proc=Delta, protocol=LoihiProtocol)
+@requires(CPU)
+@tag('floating_pt')
+class PyDeltaModelFloat(AbstractDeltaModel):
```
mgkwill, line 194: Same as above, duplicate code that could be put into a base class for PyDeltaModelFloat and PyDeltaModelFixed.

src/lava/proc/sdn/models.py
```diff
@@ -0,0 +1,296 @@
...
+        s_out_scaled = self.delta_dynamics(a_in_data)
+        s_out = np.right_shift(s_out_scaled, self.wgt_exp)
+        self.act = a_in_data
+        self.s_out.send(s_out)
+
+
+@implements(proc=SigmaDelta, protocol=LoihiProtocol)
+@requires(CPU)
+@tag('floating_pt')
+class PySigmaDeltaModelFloat(AbstractSigmaDeltaModel):
```
mgkwill, line 249: Same as above.

mgkwill (approved):
This looks pretty good to me.

There are a few optimizations available to you. See comments.

src/lava/proc/sdn/models.py
```diff
@@ -0,0 +1,296 @@
...
+        ----------
+        act_data : np.ndarray
+            data to be encoded
+
+        Returns
+        -------
+        np.ndarray
+            delta encoded data
+        """
+        delta = act_data - self.act + self.residue
```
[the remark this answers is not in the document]
bamsumit: Thanks for pointing it out. The initialization in abstract class was missing. Fixed it.

src/lava/proc/sdn/models.py
bamsumit: Yes this is for clarity to keep track of vars that needs to be overwritten in the eventual leaf implementation.

src/lava/proc/sdn/models.py
bamsumit: I have refactored it. Thx.

src/lava/proc/sdn/models.py
```diff
@@ -0,0 +1,284 @@
...
+            Input data
+
+        Returns
+        -------
+        np.ndarray
+            decoded data
+        """
+        return a_in_data + self.sigma
+
+    def run_spk(self) -> None:
```
tim-shea, line 54: This is a quick test.

src/lava/proc/sdn/process.py
```diff
@@ -0,0 +1,151 @@
...
+    Delta dynamics (with cumulative error):
+    delta   = act_new - act + residue           # delta encoding
+    error   = error + delta                     # error accumulation
+    s_out   = delta if abs(error) > vth else 0  # spike mechanism
+    error   = error * (1 - H(s_out))            # error reset
+    resiude = delta - s_out                     # residue accumulation
+    act = act
+
+    Parameters
+    ----------
```
tim-shea, line 69: Let's add the 3 kw args params here.

src/lava/proc/sdn/process.py
```diff
@@ -0,0 +1,151 @@
...
+    act_new = act_fx(sigma + bias)              # activation
+    delta   = act_new - act + residue           # delta encoding
+    error   = error + delta                     # error accumulation
+    s_out   = delta if abs(error) > vth else 0  # spike mechanism
+    error   = error * (1 - H(s_out))            # error reset
+    resiude = delta - s_out                     # residue accumulation
+    act = act
+
+    Parameters
+    ----------
```
tim-shea, line 121: Same as above.

Issue #181, opened 2022-02-15
Create NxSDKRuntimeService


mgkwill:
Objective of issue: Add NxSDKRuntimeService to runtime_services.py to enable running on Loihi

**Lava version:**
- [x] **0.3.0** (feature release)

**I'm submitting a ...** 
- [x] feature request

**Current behavior:**

- NxSDKRuntimeService does not exist

**Expected behavior:**

- NxSDKRuntimeService is able to interact with Loihi through NxCore.

PR #182, opened 2022-02-15
Add NxSDKRuntimeService

mgkwill:
Issue Number: #181

Objective of pull request: Create NxSDKRuntimeService

## Pull request checklist

## Pull request type

Please check your PR type:
- [x] Feature

- NxSDKRuntimeService does not exist and Lava cannot interact with Loihi.

## What is the new behavior?
- NxSDKRuntimeService is able to interact with Loihi through NxCore.

- [x] No


src/lava/magma/core/model/nc/model.py
```diff
@@ -38,3 +102,146 @@ def allocate(self, net: Net):
...
+        super(AbstractNcProcessModel, self).__init__(proc_params,
+                                                     loglevel=loglevel)
+        self.board = board
+
+    def start(self):
+        self.service_to_process.start()
+        self.process_to_service.start()
+        for p in self.nc_ports:
+            p.start()
+        # self.board.start()
```
mgkwill, line 120: Remove self.board.start()

mgkwill:
Add docstring everywhere.

mgkwill:
NxSDKRuntimeService does not meet class naming standards. 

Change to NxSdkRuntimeService.

Same everywhere for NxSDK.

src/lava/magma/core/model/nc/model.py
```diff
@@ -38,3 +106,148 @@ def allocate(self, net: Net):
...
+                    self.board.pause()
+                    self.process_to_service.send(MGMT_RESPONSE.PAUSED)
+                    self.join()
+                    return
+                try:
+                    if enum_equal(cmd, MGMT_COMMAND.RUN):
+                        self.process_to_service.send(MGMT_RESPONSE.DONE)
+                        num_steps = self.service_to_process.recv()
+                        if num_steps > 0:
+                            # self.board.run(numSteps=num_steps, aSync=False)
```
mgkwill, line 158: Uncomment now the unit tests are skipped with condition.

src/lava/magma/core/model/nc/model.py
```diff
@@ -38,3 +106,148 @@ def allocate(self, net: Net):
...
         pass
+
+
+class NcProcessModel(AbstractNcProcessModel):
+    def __init__(self, proc_params: ty.Dict[str, ty.Any],
+                 board: ty.Type[NxBoard],
+                 loglevel=logging.WARNING):
+        super(AbstractNcProcessModel, self).__init__(proc_params,
+                                                     loglevel=loglevel)
+        self.board = board
```
joyeshmishra, line 117: I understand this is probably mock code at the moment. However, just mentioning that NcProccessModel will work with the Net Interface and should not have the board object. Please remove it from here.

src/lava/magma/core/model/nc/model.py
```diff
@@ -38,3 +106,148 @@ def allocate(self, net: Net):
...
+        service is informed about completion. The loop ends when the
+        STOP command is received."""
+        selector = CspSelector()
+        channel_actions = [(self.service_to_process, lambda: 'cmd')]
+        action = 'cmd'
+        while True:
+            if action == 'cmd':
+                cmd = self.service_to_process.recv()
+                if enum_equal(cmd, MGMT_COMMAND.STOP):
+                    self.board.stop()
```
joyeshmishra, line 144: There might be multiple NcProcessModels. As every process model will get mgmt command and every one tries to relay it to board - wouldn't this be a problem. I would suggest moving all the board mgmt logic to NxSDKRuntimeService.

tests/lava/magma/runtime/test_runtime_service.py
```diff
@@ -1,14 +1,27 @@
+import logging
```
joyeshmishra, line 1: Is there an unit test from the runtime level - i.e. if there is a single process model which is NeuroCore (simple for now) and you add that and then you call process.run() - If you need something in the compiler, you can mock it. But this test would demonstrate if the runtime -> NxSDKRuntimeService -> process_model flow is working.

If you want more clarity on the test, let's schedule a 30 min and we can quickly write up the unit test.

joyeshmishra (changes requested):

src/lava/magma/core/model/nc/model.py
mgkwill: ACK. That was not entirely clear to me.

src/lava/magma/core/model/nc/model.py
mgkwill: > There might be multiple NcProcessModels. As every process model will get mgmt command and every one tries to relay it to board - wouldn't this be a problem. I would suggest moving all the board mgmt logic to NxSDKRuntimeService.

ACK, no problem. I asked about this in a chat with you but perhaps I misunderstood or didn't communicate what I meant:
"[2/11/2022 4:44 PM] Marcus
Is it a problem if all of the process models have a reference to the same board?"
[2/11/2022 4:45 PM] Joyesh
They must have reference to the same board. There is a single board at the moment.
I mean there is no problem"

tests/lava/magma/runtime/test_runtime_service.py
mgkwill: I have the start of a unit test for "runtime -> NxSDKRuntimeService -> process_model" at bottom and in test_runtime_service.py. Since it was a task that seemed to require working channels between runtime and NxSDKRuntimeService I am working on it this week as part sprint work on runtime to runtimeservice channel communication.

In the unit test I was deadlocking on `cmd = self.service_to_process.recv()` in lava/magma/core/model/nc/model.py
and it seemed there were two problems. One I was not ACKing for each message. I believe I fixed this. Two I did not start up the runtime and thus NxSDKRuntimeService and NcProcModel were waiting indefinitely for a message to come from runtime.

On the second point I could use some help.

```
    # @unittest.skipUnless(run_loihi_tests, \
    #                      "runtime_to_runtime_service_to_process \
    #                      communication")
    @unittest.skip("runtime_to_runtime_service_to_process_to_loihi \
                   communication")
    def test_runtime_service_loihi_channels_start_run(self):
        node: Node = Node(HeadNode, [])
        exec: Executable = Executable()
        exec.node_configs.append(NodeConfig([node]))
        runtime: Runtime = Runtime(exec, ActorType.Mult
```

src/lava/magma/runtime/runtime_services/interfaces.py
```diff
@@ -0,0 +1,48 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+# See: https://spdx.org/licenses/
+import typing as ty
+from abc import ABC, abstractmethod
+
+from lava.magma.compiler.channels.pypychannel import CspRecvPort, CspSendPort
+from lava.magma.core.sync.protocol import AbstractSyncProtocol
+
+
+class AbstractRuntimeService(ABC):
```
awintel, line 11: Missing docstring. What is a RuntimeService? What is it's relation to the Runtime and ProcModels?

src/lava/magma/runtime/runtime_services/runtime_service.py
```diff
@@ -295,3 +261,103 @@ def run(self):
                     if not enum_equal(rsp, MGMT_RESPONSE.DONE):
                         raise ValueError(f"Wrong Response Received : {rsp}")
                 self.service_to_runtime.send(MGMT_RESPONSE.DONE)
+
+
+class NxSDKRuntimeService(NcRuntimeService):
+    """NxSDK RuntimeService that implements NxCore SyncProtocol.
```
awintel, line 268: There is no NxCore SyncProtocol. Should be rewritten after AbstractRuntimeService docstring is written.
A RuntimeService is a service process coordinating communication and execution within one SyncDomain implementing a particular SyncProtocol for all connected ProcModels.
Just like the PyLoihiRuntimeService, this one is also implementing the Loihi protocol. NxCore is just an API to configure and manage a Loihi system.

awintel: Seems like there is lots of code duplication in this class with other RuntimeServices.

src/lava/magma/runtime/runtime_services/runtime_service.py
```diff
@@ -295,3 +261,103 @@ def run(self):
...
+                    self.service_to_runtime.send(MGMT_RESPONSE.ERROR)
+
+                    self._send_pm_cmd(MGMT_COMMAND.STOP)
+                    return
+
+    def get_board(self) -> NxBoard:
+        if self.board is not None:
+            return self.board
+        else:
+            AssertionError("Cannot return board, self.board is None")
```
awintel, line 369: Not a very meaningful assertion. There's no obvious reason why you could not return board=None. 
What's the actual reason this has to fail? Perhaps because the system has not been booted or set up yet?

src/lava/magma/core/model/nc/ports.py
```diff
@@ -0,0 +1,358 @@
+# Copyright (C) 2022 Intel Corporation
```
awintel, line 1: Does all the code duplication in this file motivate a refactoring of other classes if this is really necessary?
Perhaps file an issue about that, because clearly that should not remain that way.

awintel: It's also not quite clear why we would need many of these things in here because clearly we don't need Python implementation of recv() fr a NcProcModel executing on a NeuroCore. 
We only needed that for PyProcModels and corresponding PyPorts executing as Python code on CPU.

src/lava/magma/core/model/nc/model.py
```diff
@@ -38,3 +102,146 @@ def allocate(self, net: Net):
         Note: This should work as before.
         """
         pass
+
+
+class NcProcessModel(AbstractNcProcessModel):
```
awintel, line 111: Unclear what this class is for. An NcProcModel is for allocating resources on a NeuroCore. Execution of that ProcModel means some neuro cores will run according to the resource configurations requested by this ProcModel. Why do we need a Python implementation for running an NcProcModel with all kinds of channel management?

The real NeuroCoreProcess running on a neuro core will communicate with NxRuntime for pausing and stopping. If any SeqProc wants to interact with Vars, it will do so directly. This does not have to go through the NxSdkRuntimeService. If the user wants to manipulate Vars at runtime, the most natural way would be for the NxSDKRuntimeservice to do this via the Board directly.

awintel: Seems like there is overall architectural documentation for the overall usage modal and why this is necessary.

awintel (commented):
Before approving or requesting changes I need some fundamental architectural questions clarified.

src/lava/magma/runtime/runtime_services/interfaces.py
mgkwill: Will be resolved in another PR via #184, as this would be useful for upcoming release that this change is not a part of.

joyeshmishra (approved):

ysingh7 (approved):

Issue #186, opened 2022-02-17
Fix intermittent test_explicit_Ref_Var_port_write on MacOs


mgkwill:
Objective of issue: Fix intermittent test_explicit_Ref_Var_port_write on MacOs

**Lava version:**
- [x] **0.2.1** (bug fixes)

**I'm submitting a ...** 
- [x] bug report

**Current behavior:**
- Intermittent failure of MacOS unit test runs: 
```
======================================================================
FAIL: test_explicit_Ref_Var_port_write (tests.lava.magma.runtime.test_ref_var_ports.TestRefVarPorts)
Tests the connection of a RefPort to an explicitly created VarPort.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/runner/work/lava/lava/tests/lava/magma/runtime/test_ref_var_ports.py", line [120](https://github.com/lava-nc/lava/runs/5238112313?check_suite_focus=true#step:5:120), in test_explicit_Ref_Var_port_write
    self.assertTrue(np.all(recv.var2.get() == np.array([7., 7., 7.])))
AssertionError: False is not true
```

**Expected behavior:**
- There is not an Intermittent failure of MacOS unit test runs.

**Steps to reproduce:**
- Run unit tests in CI a few times. See: https://github.com/lava-nc/lava/runs/5238112313?check_suite_focus=true#step:5:114

mgkwill:
This may not be an issue with Ref Var code but some underlying issue with lava, it seems like a race issue with port communication as I don't get this as much when using verbose `python -m unittest -v`

Saw this again in the CI today:
```
======================================================================
FAIL: test_implicit_Ref_Var_port_write (tests.lava.magma.runtime.test_ref_var_ports.TestRefVarPorts)
Tests the connection of a RefPort to an implicitly created VarPort.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/runner/work/lava/lava/tests/lava/magma/runtime/test_ref_var_ports.py", line 206, in test_implicit_Ref_Var_port_write
    self.assertTrue(np.all(recv.var2.get() == np.array([9., 9., 9.])))
AssertionError: False is not true

----------------------------------------------------------------------
Ran 183 tests in 83.294s

FAILED (failures=1, skipped=9)
```

PhilippPlank:
Yes, I do not think this is related to MacOS. The question is what the reason for the failure is, as it is not reproduceable locally for me.

mgkwill:
[...]

PR #190, opened 2022-02-24
Fixing priority of channel commands in model

PhilippPlank:
Issue Number: #186 

Objective of pull request: Fixing errors in unit tests with Ref-/VarPort interaction

## Pull request checklist

## Pull request type

 

 

Please check your PR type:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] Documentation changes

- Intermittent failure of unit test runs (see Issue).

## What is the new behavior?

- There is not an Intermittent failure of unit test runs.

- [ ] Yes
- [x] No


mgkwill:
Thanks for submitting this @PhilippPlank!

src/lava/magma/core/model/py/model.py
```diff
@@ -419,7 +419,8 @@ def add_ports_for_polling(self):
                     if isinstance(csp_port, CspRecvPort):
                         def func(fvar_port=var_port):
                             return lambda: fvar_port
-                        self._channel_actions.append((csp_port, func(var_port)))
+                        self._channel_actions.insert(0,
```
mgkwill, line 422: So rather than appending to the `_channel_actions` list 
```
self._channel_actions: ty.List[ty.Tuple[ty.Union[CspSendPort, CspRecvPort], ty.Callable]]
```
we are overwriting idx 0.

awintel: We are prepending an element rather than appending.

awintel (approved):
Admittedly, I don't understand the context of this issue. But the change seems minimal. So if it works - great.

src/lava/magma/core/model/py/model.py
PhilippPlank: Yes, insert is not overwriting existing elements.

PhilippPlank:
> Admittedly, I don't understand the context of this issue. But the change seems minimal. So if it works - great.

To give context. If we have VarPorts in a ProcessModel and we are in management phase, the channel action of each VarPort gets added to the list of possible channel actions to listen if anyone sent something to a VarPort.

In this unit test the list of channel actions is [default "cmd" channel, first_var_port, second_var_port]

This list is then given to the select function, which picks an action as soon as something arrives. 

The problem now was, that when entering the management phase we had data on both VarPort, select picks the first one to service. After that the loop starts from the beginning and it services the first VarPort.
As this takes time, the runtime_service already sent a STOP command, since it is also the last configured time step.

Now in some cases the STOP command was sent, before we reach the select again after being done with the first VarPort. Now the select has 2 channel actions to pick from, and picked the default "cmd" channel, since it comes first in the list. This stopped the execution before we had a chance to service the second VarPort and hence the unit test failed.

Solution: order the channel command list so that VarPorts come first, so we can only listen to other commands when they all have been serviced (if needed).

mgkwill:
Ah yes. Prepending rather than replacing. 

OK. Makes sense to me.

mgkwill (approved):

src/lava/magma/core/model/py/model.py
mgkwill: My mistake. 👍🏾 

Appending.

Issue #204, opened 2022-03-01
RefPort's sometimes handled a time step late


PhilippPlank:
Objective of issue: Deterministic behavior of when data sent via a RefPort is received by the target process. Occasionally it was possible that the receiving process moved to the next phase before all of its VarPorts got serviced, because sending data via a RefPort does not block. Introducing of a wait() method for RefPorts makes sure that sent data is serviced before the process finishes (which blocks advancing a time step and ensures all VarPorts on the receiving process will be serviced beforehand).

**Lava version:**
- [ ] **0.3.0** (feature release)
- [ ] **0.2.1** (bug fixes)
- [x] **0.2.0** (current version)
- [ ] **0.1.2**

**I'm submitting a ...** 

- [ ] bug report
- [ ] feature request
- [ ] documentation request

**Current behavior:**

- Occasionally it was possible that the receiving process moved to the next phase before all of its VarPorts got serviced, because sending data via a RefPort does not block. 

**Expected behavior:**

- 

**Steps to reproduce:**

- Introducing of a wait() method for RefPorts makes sure that sent data is serviced before the process finishes (which blocks advancing a time step and ensures all VarPorts on the receiving process will be serviced beforehand).

**Related code:**

```
insert short code snippets here
```

**Other information:**

```
insert the output from lava debug here
```

PR #205, opened 2022-03-01
RefPort's sometimes handled a time step late

PhilippPlank:
Issue Number: #204 

Objective of pull request: Deterministic behavior of when data sent via a RefPort is received by the target process. Occasionally it was possible that the receiving process moved to the next phase before all of its VarPorts got serviced, because sending data via a RefPort does not block. Introducing of a wait() method for RefPorts makes sure that sent data is serviced before the process finishes (which blocks advancing a time step and ensures all VarPorts on the receiving process will be serviced beforehand).

## Pull request checklist

## Pull request type

 

 

Please check your PR type:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] Documentation changes

- Occasionally it was possible that the receiving process moved to the next phase before all of its VarPorts got serviced, because sending data via a RefPort does not block.

## What is the new behavior?

- Introducing of a wait() method for RefPorts makes sure that sent data is serviced before the process finishes (which blocks advancing a time step and ensures all VarPorts on the receiving process will be serviced beforehand).

- [ ] Yes
- [x] No


src/lava/magma/core/model/py/ports.py
```diff
@@ -513,6 +513,14 @@ def write(
         """
         pass
 
+    # TODO: (PP) This should be optimized by a proper CSPSendPort wait
+    def wait(self):
+        """Method which block execution until (all) previous write requests have
```
awintel, line 518: This is a high-level user facing method. I suggest to write a proper docstring either at the RefPort or method level that explains what this method does and why the RefPort needs it.

In a final comment, I would state that this is a preliminary implementation that achieves the desired blocking behavior until the earlier WRITE has been acknowledged via a read(). This is inefficient and will be optimized later at the CspChannel level.

awintel (changes requested):
Looks fine. But you should update the docstring.

src/lava/magma/core/model/py/ports.py
```diff
@@ -513,6 +513,14 @@ def write(
...
+         been placed at the receiver. Currently, achieved by a simple read(),
+         which ensures the writes have been finished. """
+
+        self.read()
+
```
bamsumit, line 523: Doesn't it make sense to call self.wait() at the end of self.write(...) routine? Is there any reason why we would want the non-deterministic behavior?

src/lava/proc/io/reset.py
```diff
@@ -59,6 +59,7 @@ def post_guard(self) -> None:
 
     def run_post_mgmt(self) -> None:
         self.state.write(0 * self.state.read() + self.reset_value)
```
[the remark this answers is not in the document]
bamsumit: We also need to add self.state.wait() in the dataloader as well if we want to not always call it in the write() method. I'll create a separate PR for it.

bamsumit (approved):

bamsumit:
Tested it out on PN SNN notebook. The the results are consistent.

src/lava/proc/io/reset.py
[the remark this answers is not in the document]
PhilippPlank: I removed the read, since the port knows its shape (although only as protected member) and it seems that is the only reason for the read here.

src/lava/magma/core/model/py/ports.py
```diff
@@ -516,9 +516,19 @@ def write(
     # TODO: (PP) This should be optimized by a proper CSPSendPort wait
     def wait(self):
         """Method which block execution until (all) previous write requests have
```
awintel, line 521: No need to say that this is a method. 

"Blocks execution until receipt of prior 'write' commands (sent from RefPort to VarPort) have been acknowledged..."

awintel (approved):
I would still recommend to improve the docstring. 
The fundamental behavior of the RefPort should be explained in the class docstring.

Approving anyways since this is just a style issue that can be fixed later. Just don't forget about it.

src/lava/magma/core/model/py/ports.py
PhilippPlank: Done

Issue #228, opened 2022-03-14
Revision of Monitor API 


elvinhajizada:
### User story
As a user, I would like to the API of `Monitor` process to be more intuitive and simple.

### Conditions of satisfaction
- Data extraction from the monitors should be more straightforward.
- API allows to probe with some time offset and interval.
- API allows multi-var probing with one 1-2 lines of code
- The test coverage of the newly added module is >95%.
- The code has been reviewed and is merged into `main`.

### Acceptance tests
- Test that revised API is compatible with multi-var probing feature
- Test that data collected with time offset and interval is as expected.
- Test that this is faster than creating one Monitor per var

### Additional reference info based on original issue (remove once this user story is finalized):
**Current behavior:**

- There is only one Monitor class, 
- That can monitor only one Var or OutPort of a single Process
- User need to use get_data() function to create a data structure of collected data, which then can only be access by process name and Var/OutPort name combination (string) .e.g.
`data1 = monitor1.get_data()`
`data2 = monitor2.get_data()`
`volt_data = data1[neuron.name][neuron.v.name]`
`spike_data = data2[neuron.name][neuron.s_out.name]`

**Expected behavior:**

-  In the new design we will have multiple classes (e.g. Recorder, Live Streamer, etc.) inherited from an abstract Monitor class.
- `get_data()` function should accept the Var or OutPort and directly return the data 
`data_v = recorder.get_data(neuron1.v)`
- Monitors will be able to probe with some time offset and interval with the following API:
`recorder.probe(neuron.v, interval=2, offset=0)`
- Multiple Var probing should be done by roving list of Vars to the `probe()` funtion:
`recorder.probe([neuron1.v, neuron2.v])`

PR #229, opened 2022-03-14
Nc monitor

elvinhajizada:
Issue Number: #228 

Objective of pull request: To provide unittests for demonstrating updated API for new Monitor design

## Pull request checklist

- [X] [Issue](https://github.com/lava-nc/lava/issues) created that explains the change and why it's needed
- [X] Tests are part of the PR (for bug fixes / features)
- [X] PR conforms to [Coding Conventions](https://lava-nc.org/developer_guide.html#coding-conventions)
- [X] [PR applys BSD 3-clause or LGPL2.1+ Licenses](https://lava-nc.org/developer_guide.html#add-a-license) to all code files

## Pull request type

 

 

Please check your PR type:

- [ ] Bugfix
- [ ] Feature
- [X] Code style update (formatting, renaming)
- [ ] Build related changes
- [ ] Documentation changes

- There are unittests for current version of the Monitor class

## What is the new behavior?

- New (but failing) unittests for Monitor class are added to showcase and collect feedback for new API. Corresponding implementation of the new design is not here yet

- [X] Yes
- [ ] No

This PR is a draft and is not meant to merged yet, rather to collect feedback for new API of Monitor


tests/lava/proc/monitor/test_monitors.py
```diff
@@ -342,6 +342,141 @@ def test_proc_params_accessible_in_proc_model(self):
         self.assertEqual(next(iter(exe.py_builders)).proc_params,
                          monitor.proc_params)
 
+    # New API unittests
+    def test_single_recorder_with_single_targer(self):
```
mathisrichter, line 346: targer -> target

tests/lava/proc/monitor/test_monitors.py
```diff
@@ -342,6 +342,141 @@ def test_proc_params_accessible_in_proc_model(self):
...
+        """Check if a single Recorder can monitor the spike output of a
+        neuron"""
```
mathisrichter, line 348: missing full stop after "neuron"

tests/lava/proc/monitor/test_monitors.py
```diff
@@ -342,6 +342,141 @@ def test_proc_params_accessible_in_proc_model(self):
...
+        shape = (1,)
+        num_steps = 6
+        neuron = LIF(shape=shape,
+                     vth=3,
+                     bias=1)
+
+        recorder = Recorder()
+        recorder.probe(neuron.s_out)
```
mathisrichter, line 356: I'd still prefer 'monitor' as the method name. What do you think? I don't think we should stick to the "neuron" analogy with a probe and a cell as Lava is more general. I know Andreas likes "probe" better but maybe we can convince him.

tests/lava/proc/monitor/test_monitors.py
mathisrichter, line 346: The name should reflect that you are monitoring a Port. There should then be another unit test for monitoring an individual Var.

tests/lava/proc/monitor/test_monitors.py
```diff
@@ -342,6 +342,141 @@ def test_proc_params_accessible_in_proc_model(self):
...
+
+        neuron.run(condition=RunSteps(num_steps=num_steps),
+                   run_cfg=Loihi1SimCfg())
+
+        spike_data = recorder.get_data(neuron.s_out)
+        self.assertTrue(np.all(spike_data == np.array([[0, 0, 1, 0, 0, 1]]).T))
+
+    def test_single_recorder_with_multiple_targets_from_single_proc(self):
+        """Check if a single Recorder can monitor multiples targets,
```
mathisrichter, line 365: multiples -> multiple

tests/lava/proc/monitor/test_monitors.py
```diff
@@ -342,6 +342,141 @@ def test_proc_params_accessible_in_proc_model(self):
...
+        here the spike output and voltage of a neuron"""
+        shape = (1,)
+        num_steps = 6
+        neuron = LIF(shape=shape,
+                     vth=3,
+                     bias=1)
+
+        recorder = Recorder()
+        recorder.probe([neuron.s_out, neuron.v])
```
mathisrichter, line 374: I think we decided to not allow this anymore. There is an issue with specifying multiple sources and then also being able to specify `interval` and `offset` for them. Unless we are only supporting this kind of method call with multiple sources that all have the same `interval` and `offset`? If so, we should specify that here, maybe even put it in a comment.

tests/lava/proc/monitor/test_monitors.py
```diff
@@ -342,6 +342,141 @@ def test_proc_params_accessible_in_proc_model(self):
...
+                   run_cfg=Loihi1SimCfg())
+
+        spike_data = recorder.get_data(neuron.s_out)
+        voltage_data = recorder.get_data(neuron.v)
+        self.assertTrue(np.all(spike_data == np.array([[0, 0, 1, 0, 0, 1]]).T))
+        self.assertTrue(
+            np.all(voltage_data == np.array([[1, 2, 0, 1, 2, 0]]).T))
+
+    def test_single_recorder_with_targets_from_multiple_procs(self):
+        """Check if a single Recorder can monitor multiples targets,
```
mathisrichter, line 386: multiples -> multiple

mathisrichter (changes requested):
Overall, we may not yet need these full end-to-end unit tests that run everything and then check the results. A more fine grained approach that we will have to do anyway is to unit test the API and check, for instance, only that the required Ports are instantiated in the Process. Later, we will have to add unit tests to the compiler, checking that it does the right thing with a mocked version of whatever it would receive from a network with a Recorder. Then do the same thing for the builder. And then, in the end, we would probably write a few end-to-end tests that are really required.
That approach also helps you write passing unit tests from the start, even though you are only adding functionality iteratively.

tests/lava/proc/monitor/test_monitors.py
elvinhajizada: Yes I agree we don't need allow multi-source monitoring, it complicates the implementation unnecessarily.

elvinhajizada:
@mathisrichter I have some of fine-grained tests for old API, they just need to be adapted to the new API

Issue #232, opened 2022-03-21
LavaProfiler User Interface


PhilippPlank:
Objective of issue: Implement an universal tool to capture power, energy and execution time in Lava which works on all backends with one interface across backends and which gives insights about power and performance of the components. This tool will be the LavaProfiler.

**Lava version:**

- [ ] **0.4.0** (feature release)
- [ ] **0.3.1** (bug fixes)
- [x] **0.3.0** (current version)
- [ ] **0.2.0**
- [ ] **0.1.2**

**I'm submitting a ...** 

- [ ] bug report
- [x] feature request
- [ ] documentation request

**Current behavior:**

- No LavaProfiler implemented.

**Expected behavior:**

- LavaProfiler allows users to capture power and performance metrics.

**Steps to reproduce:**

- NA

**Related code:**

```
insert short code snippets here
```

**Other information:**

```
insert the output from lava debug here
```

PR #233, opened 2022-03-21
Initial version of LavaProfiler user interface

PhilippPlank:
Issue Number: #232 

Objective of pull request: Implement an universal tool to capture power, energy and execution time in Lava which works on all backends with one interface across backends and which gives insights about power and performance of the components. This tool will be the LavaProfiler.

## Pull request checklist

## Pull request type

 

 

Please check your PR type:

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] Documentation changes

- No LavaProfiler

## What is the new behavior?

- LavaProfiler user interface defined and sketch of the main methods in the profiler class.

- [ ] Yes
- [ ] No


src/lava/utils/profiler.py
```diff
@@ -22,7 +22,8 @@
 from lava.magma.runtime.runtime import Runtime
 from lava.magma.core.process.message_interface_enum import ActorType
 from lava.magma.compiler.compiler import Compiler
-from lava.magma.core.resources import Loihi1NeuroCore, Loihi2NeuroCore
+from lava.magma.core.resources import (
+    AbstractComputeResource, Loihi1NeuroCore, Loihi2NeuroCore)
 
 
 class LavaProfiler(ABC):
```
awintel, line 29: Why LavaProfiler? Profiler should be enough.

Why does this have to be an ABC? In this case I would probably call it AbstractProfiler.

awintel, line 29: Always good to have some docstrings by the time of a PR in which you expect feedback from others.

src/lava/utils/profiler.py
```diff
@@ -34,7 +35,7 @@ def __init__(self, start: int = 0, end: int = 0,
         self.end = end
         self.bin_size = bin_size
         self.buffer_size = buffer_size
-        self.loihi_flag = False  # Exchange with list of resources
+        self.used_resources: ty.List[AbstractComputeResource] = []
 
     def profile(self, proc: AbstractProcess):
```
awintel, line 40: docstring

src/lava/utils/profiler.py
```diff
@@ -12,3 +12,122 @@
...
+            # 1. get proc_map
+            # proc_map = initializer._map_proc_to_model(proc, run_cfg)
+            proc_map = compiler._map_proc_to_model(
+                compiler._find_processes(proc), run_cfg)
+
+            # 2. modify proc_map
+            proc_map = self._modify_proc_map(proc_map)
+
+            # 3. set up profiler
+            self._set_profiler(proc_map)
```
awintel, line 94: 'set up' or 'set'?

tests/lava/utils/test_profiler.py
```diff
@@ -0,0 +1,73 @@
...
+class MyRunCfg(RunConfig):
+    def select(self, proc, proc_models):
+        return proc_models[0]
+
+
+class TestLavaProfiler(unittest.TestCase):
+    def test_init(self):
+        """TBD"""
+        proc = P()
+        profiler = LavaProfiler()
```
awintel: I thought this is an abstract class?

awintel (changes requested):
Good start but lots to do left...

src/lava/utils/profiler.py
```diff
@@ -12,3 +12,125 @@
 elementary operation has a defined execution time and energy cost, which is
 used in a performance model to calculate execution time and energy.
 """
```
phstratmann, line 14: Docstring: 
will contain -> contains [otherwise, this needs to be rewritten once it works next month, and we will forget it]
power and performance of workloads -> isn't power an aspect of performance?

Maybe a few lines on what the Profiler simulation is good for? 
- Simulate a network on CPU before it is being deployed - in which situations is that feature helpful? If people don't have access to Loihi yet. If they want to understand the performance on a fine-granular layer. ...

src/lava/utils/profiler.py
```diff
@@ -12,3 +12,125 @@
...
+            proc._runtime.initialize()
+
+        proc._runtime.start(condition)
+
+    def _modify_proc_map(self, proc_map):
+        """Check if chosen process models have a profileable version and
+        exchange the process models accordingly.
+        Tell the user which Processes will not be profiled, as they lack a
+        profileable ProcModel."""
+        ...
```
phstratmann, line 115: Should we automatically switch process models? I could imagine that people may get confused in some cases. Let's assume the user first runs a process without profiler and the compiler chooses a process model that is not profilable. Then the user runs the same process with profiler. The profiler will automatically switch the process model to one that can be profiled. But these process models may differ - maybe because of a bug, maybe because of other reasons. The user will not have expected to see any different process behavior just because (s)he activated the profiler. Instead, I would expect that I receive an error message if the default process model cannot be profiled.

src/lava/utils/profiler.py
```diff
@@ -12,3 +12,122 @@
...
+
+    def __init__(self, start: int = 0, end: int = 0,
+                 bin_size: int = 1, buffer_size: int = 1000):
+        self.start = start
+        self.end = end
+        self.bin_size = bin_size
+        self.buffer_size = buffer_size
+        self.loihi_flag = False  # Exchange with list of resources
+
+    def profile(self, proc: AbstractProcess):
```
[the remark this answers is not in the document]
phstratmann: I still think that the user API would be better if the user could simply define a profiler without having to change his line
proc.run(...),
i.e. the interface that PP coded here. But I also see Andreas point: Overwriting the run function feels bad. In particular, what happens if we ever change the general run method of Processes? Then we may forget to change it in the Profiler, and it will stop working.
What does the Profiler actually do under the hood? Does it just count how often the internal run_spk method is called, and how often spikes are being sent? In that case, could we 
1) ask people to just add a decorator @count to count runs in front of the run_spk function:
class MyProcModel(list):
    @counter
    def run_spk(self, *args, **kwargs):
        ...
with 
def counter(f):
    def count(*args, **kwargs):
        count.calls += 1
        return f(*args, **kwargs)
    count.calls = 0
    return count
This decorator would mean that we don't need to rewrite the whole run method.
2) subscribe to the port of the Process to read its sent spikes?

To me, that would be way less invasive. But I know too little about what's going on under the hood to say if anything like that would work.

src/lava/utils/profiler.py
```diff
@@ -12,3 +12,125 @@
...
+        self.used_resources: ty.List[AbstractComputeResource] = []
+
+    def profile(self, proc: AbstractProcess):
+        proc.run = types.MethodType(self.run, proc)
+
+    def get_energy(self) -> np.array:
+        """Returns the energy estimate per time step in µJ."""
+        ...
+        return 0
+
```
phstratmann, line 47: Should we give the user the option to either receive the whole time series or just the total time / energy? I could imagine that for particularly long runs, we run into memory or runtime problems if we store one value for each time step. If we just accumulate the values, it may often suffice.
In addition, what's about a value to provide the frequency of measurements? We may often not need a value each time step, but only each 1000th time step.

phstratmann (changes requested):
Great work! Some comments and questions, though.

src/lava/utils/profiler.py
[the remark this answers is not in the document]
PhilippPlank: Both versions are slide by slide in the PowerPoint ;)

@phstratmann We need to do quite a few things under the hood, especially modifying the compilation process (the sketch of what we need to do is in form of comments and mock methods in this PR).
We do not really count how often run_spk is called.

src/lava/utils/profiler.py
PhilippPlank: I moved this Docstring to the Profiler class docstring without the "will".

src/lava/utils/profiler.py
PhilippPlank: My approach would be that each ProcModel has a profileable counter part ProcModel, which only adds the operation counters.
So for the LIF neuron for example, we would need a profilable ProcModel for "fixed_pt" and "floating_pt" and there is a 1:1 mapping to exchange them if the profiler is present.

If there is no such ProcModel, then the user will be informed that this Process is not considered by the Profiler. If no chosen ProcModel has a profileable version and we run on simulation only, than there will be an error stating no Process is considered for the Profiler.

src/lava/utils/profiler.py
PhilippPlank: Sure, this is just a first example of getting results. We can offer more sophisticated options, including standard plots etc.

src/lava/utils/profiler.py
awintel: What does this mean in terms of code duplication? Does it mean there is a ProcModel class and then an almost identical ProcModel class that just adds counters?

Before we go into a whole lot of implementation, as usual, we should first write down an end to end (mock) example of what we are trying to enable and then agree that this is the best way to go. Such decisions are best made not in the abstract, for people who have not thought about the pros and cons deeply before, but using a concrete exmple.

Do we have such an example already?

If not I suggest you draft one and share it. We are at an important fork in the road, so we should get some wider input.

PhilippPlank: I added an example in proc/lif/models.py

We can inherit the ProcModel and add the code for operation counters.

tim-shea:
@PhilippPlank I propose to close this unmerged as it has grown too far out of sync with our current plan for PnP, are you okay with that?

Issue #148, opened 2021-12-06
Can't run tutorial in notebook


matham:
Objective of issue:  Tutorial crashes

**Lava version:**
- [x] current main (dec 6)
- [ ] **0.3.0** (feature release)
- [ ] **0.2.1** (bug fixes)
- [ ] **0.2.0** (current version)
- [ ] **0.1.2**

**I'm submitting a ...** 

- [x] bug report
- [ ] feature request
- [ ] documentation request

I'm simply running through the cells of the tutorial one after the other from the GUI for `tutorial02_processes` and it freezes and the logs show the following error:

```
$ jupyter notebook
[I 14:59:09.234 NotebookApp] Serving notebooks from local directory: G:\Python\lava\lava
[I 14:59:09.234 NotebookApp] Jupyter Notebook 6.4.6 is running at:
[I 14:59:09.234 NotebookApp] http://localhost:8888/
[I 14:59:09.234 NotebookApp]  or http://127.0.0.1:8888/?token
[I 14:59:09.234 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[C 14:59:09.303 NotebookApp]

    To access the notebook, open this file in a browser:
        file:///C:/Users/Matthew%20Einhorn/AppData/Roaming/jupyter/runtime/nbserver-17496-open.html
    Or copy and paste one of these URLs:
        http://localhost:8888/?token
     or http://127.0.0.1:8888/?token
[W 14:59:34.725 NotebookApp] Notebook tutorials/in_depth/tutorial02_processes.ipynb is not trusted
[I 14:59:36.725 NotebookApp] Kernel started: dc72d164-3797-45f8-b4b4-ba468a763ba4, name: python3
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Users\Matthew Einhorn\AppData\Local\Programs\Python\Python39\lib\multiprocessing\spawn.py", line 116, in spawn_main
    exitcode = _main(fd, parent_sentinel)
  File "C:\Users\Matthew Einhorn\AppData\Local\Programs\Python\Python39\lib\multiprocessing\spawn.py", line 126, in _main
    self = reduction.pickle.load(from_parent)
AttributeError: Can't get attribute 'PyLifModel' on <module '__main__' (built-in)>
[I 15:01:36.642 NotebookApp] Saving file at /tutorials/in_depth/tutorial02_processes.ipynb
```

matham:
[...]

Issue #95, opened 2021-11-23
Jupyter tutorials on Windows


PhilippPlank:
The usage of multiprocessing on Windows in combination with jupyter notebooks does not work very well (see links below).
Some of our tutorials define the ProcessModels within the jupyter notebook and during run, multiprocessing will try to spawn these models. This will most likely fail with the error
`AttributeError: Can't get attribute '<proc_model_name>' on <module '__main__' (built-in)> exitcode = _main(fd, parent_sentinel)`

The error will also not show in the notebook, but it will just stall.

The issue is quite known and a combination of problems of Windows and IPython, so this will not be solvable. The recommended solution is to put the code which multiprocessing will use into separate modules and import it in the notebook.

Should we change the tutorials to import all their code (we can still show it) or have a Windows warning/error for some of them?

https://medium.com/@grvsinghal/speed-up-your-python-code-using-multiprocessing-on-windows-and-jupyter-or-ipython-2714b49d6fac
https://github.com/microsoft/vscode-jupyter/issues/941

awintel:
Interestingly, this seems to work for some people... What is the difference that makes it work for some? It would be very instructive to keep the PyProcModels in the tutorials.

It does not work for me, but it must have worked for those that created some of the current executable tutorials. Were these never tested on Windows?

PhilippPlank:
This issue is still present in the 0.4.0 release. We are aware of it and testing different fixes, e.g. automatically converting the jupyter notebook to a python script and import the classes from the script, during execution. Such workarounds do solve the issue, but introduce additional code in the juypter tutorial itself.

If anyone has a better solution or idea, feel free to contribute or discuss!

PS: In the longer term perspective we probably move away from the multiprocessing library, which should also solve this issue.

tim-shea:
@PhilippPlank What is the status of this issue?

PR #259, opened 2022-06-14
Enable tutorials on Windows

PhilippPlank:
Issue Number: #95 and #148

Objective of pull request: All tutorials should work on Windows systems

## Pull request checklist

## Pull request type

 

 

Please check your PR type:

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] Documentation changes

- The tutorial execution stalls 

## What is the new behavior?

- The tutorial executes all cells as expected

- [ ] Yes
- [ ] No


awintel (approved):
Thanks, I'm fine with the basic mechanism for now. 
However, it would be nice to beautify it a bit more:
- Please come up with a better name for enable_win. Sounds very hack'ish. It should probably also not be in the in_depth directory if it's needed for any tutorial. Probably should go into utils?
- Please rework this statement:
"This is not part of the tutorial or Lava, but to allow this jupyter notebook to execute on Windows without errors. The class LIF and PyLifModel need to be imported from a script in order to run this tutorial on Windows due to an issue with the multiprocessing package on Windows."
-> You say it's not part of the tutorial, yet is all written within a Lava tutorial. Just give a bit of a cleaner explanation as to why such a hack is necessary and that we'll look into fixing this in the future.
-> Text says LIF AND PyLifModel need to be imported but you only import LIF
- Can you think of a way to hide all this import business? Would it be possible to just have a single function to which you pass the class of function name of a class of function defined in a tutorial. It then detects the local file it is defined in, exports it, and reimports it into the global namespace?

weidel-p:
I realized that the solution only works for tutorial 02, for the other tutorials things get more complex because Processes get overwritten.

There is some cleanup necessary, but @awintel, could you please take a look at this updated solution?

PR #279, opened 2022-07-22
Update install instructions to switch to branch.

tim-shea:
Using `git checkout v0.4.0` leads to detached HEAD state. Update to `git switch release/v0.4.0` avoids this and correctly sets local branch to track remote branch.

## Pull request type

Please check your PR type:
- [X] Documentation changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ pip install -U pip
 pip install "poetry>=1.1.13"
 git clone git@github.com:lava-nc/lava.git
 cd lava
-git checkout v0.4.0
+git switch release/v0.4.0
 ./utils/githook/install-hook.sh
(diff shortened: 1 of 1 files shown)


bamsumit (approved):

mathisrichter (dismissed):
Looks good but we first need to clarify what the status of that branch is going to be and whether we want the install instructions to rely on the branch rather than the tag.

I'll block the merge by "requesting changes" for now.

mgkwill (approved):
Yes. This is appropriate. Though only really needed if we have bug fixes that are backported.

mgkwill:
@tim-shea we should probably port this to main as well to keep in sync. Next release we will just update the version.

Issue #329, opened 2022-09-05
Bug in bit accurate ProcessModel of LIF Process 


ackurth-nc:
**Objective of issue**:
There is a bug in the bit accurate ProcessModel of the LIF Process resulting from inconsistencies in the Numpy types chosen to represent the variables.

**Lava version:**

- [ ] **0.5.0** (feature release)
- [ ] **0.4.1** (bug fixes)
- [x] **0.4.0** (current version)
- [ ] **0.3.0**
- [ ] **0.1.2**

**I'm submitting a ...** 

- [x] bug report
- [ ] feature request
- [ ] documentation request

**Current behavior:**
When setting ```bias_mant``` and ```bias_exp``` to certain (seemingly allowed) values, the ```effective_bias``` becomes negative.

**Expected behavior:**
```effective_bias``` should be positive.

**Steps to reproduce:**
```
from lava.magma.core.run_conditions import RunSteps
from lava.magma.core.run_configs import Loihi1SimCfg

from lava.proc.lif.process import LIF

run_cfg = Loihi1SimCfg(select_tag='fixed_pt')
run_condition = RunSteps(num_steps=1)

lif = LIF(shape=(1,),
          vth=2**17,
          bias_mant=2**12 - 1,
          bias_exp=5)


lif.run(run_cfg=run_cfg, condition=run_condition)
print(lif.v.get())
lif.stop()


>>> [-32]
```

PR #330, opened 2022-09-05
Fixed pt lif precision

ackurth-nc:
Issue Number: 
#329 

Objective of pull request:
There is a bug in the bit accurate ProcessModel of the LIF Process resulting from inconsistencies in the Numpy types chosen to represent the variables.
I change the data type so that it works.

## Pull request checklist

- [-] [Docs](https://github.com/lava-nc/docs) reviewed and added / updated if needed (for bug fixes / features)
- [-] [PR applys BSD 3-clause or LGPL2.1+ Licenses](https://lava-nc.org/developer_guide.html#add-a-license) to all code files

## Pull request type

Please check your PR type:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] Documentation changes

When setting ```bias_mant``` and ```bias_exp``` to certain (seemingly allowed) positive values, the ```effective_bias``` becomes negative.

## What is the new behavior?
```effective_bias``` becomes positive.

- [ ] Yes
- [x] No

## Supplemental information
Test was added.

### CODE REVIEW

File: tests/lava/proc/lif/test_models.py [hunk 1]
```diff
@@ -386,6 +386,31 @@ def test_bitacc_pm_impulse_dv(self):
         self.assertListEqual(expected_v_timeseries, lif_v)
         self.assertListEqual(expected_float_v, lif_v_float.tolist())
 
+    def test_bitacc_pm_eff_bias(self):
```
mathisrichter, line 389: Can we call this something like `test_bitacc_pm_scaling_of_bias` to make it more explicit?

tests/lava/proc/lif/test_models.py
```diff
@@ -386,6 +386,31 @@ def test_bitacc_pm_impulse_dv(self):
...
+        """
+        Tests fixed point LIF ProcessModel's scaling of threshold.
+        """
+        shape = (1,)  # a single neuron
```
mathisrichter, line 393: I think the comment may not be needed here.

mathisrichter: I would also remove the `shape` parameter and put it directly into the definition of the `LIF`, where all the other parameters are. It is not used anywhere else and is not an important detail to highlight for this test.

tests/lava/proc/lif/test_models.py
```diff
@@ -386,6 +386,31 @@ def test_bitacc_pm_impulse_dv(self):
...
+        num_steps = 1
+        # Set up high threshold and high bias current to check for potential
+        # overflow in effective bias
```
mathisrichter, line 396: Missing a full stop at the end.

tests/lava/proc/lif/test_models.py
```diff
@@ -386,6 +386,31 @@ def test_bitacc_pm_impulse_dv(self):
...
+                  vth=2 ** 17)
+
+        rcnd = RunSteps(num_steps=1)
+        rcfg = LifRunConfig(select_tag='fixed_pt')
+
+        lif.run(condition=rcnd, run_cfg=rcfg)
+        lif_v = lif.v.get()[0]
+        lif.stop()
+
+        # Check if lif_v is positive
```
mathisrichter, line 411: Missing a full stop at the end.

tests/lava/proc/lif/test_models.py
```diff
@@ -386,6 +386,31 @@ def test_bitacc_pm_impulse_dv(self):
...
+        self.assertGreater(lif_v, 0)
```
mathisrichter, line 412: For explicit bias, decay, and threshold values, you should be able to test against the exact voltage value that you expect here.

src/lava/proc/lif/models.py
```diff
@@ -69,7 +69,7 @@ class AbstractPyLifModelFixed(PyLoihiProcessModel):
     v: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=24)
     du: int = LavaPyType(int, np.uint16, precision=12)
     dv: int = LavaPyType(int, np.uint16, precision=12)
-    bias_mant: np.ndarray = LavaPyType(np.ndarray, np.int16, precision=13)
+    bias_mant: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=13)
```
mathisrichter, line 72: This is a bit unintuitive. We only "allow" 13-bit precision but need the 32 bits to do shifts later on. I know that leaving it at 16-bit is not much better...
But maybe the 32-bit variable could be introduced as a local copy of `bias_mant` inside the `scale_bias` method? That makes it more obvious what it is needed for. And I would put a comment on it to highlight that the 32 bits are important at the point.

mathisrichter (changes requested):
Nice! Only minor things to be addressed. I put in a bit more detail than I normally would.

tests/lava/proc/lif/test_models.py
ackurth-nc: Thanks for the comment, addressed in 8fa7f87.

tests/lava/proc/lif/test_models.py
ackurth-nc: Addressed in 8fa7f87.

tests/lava/proc/lif/test_models.py
ackurth-nc: Addressed in 8fa7f87.

tests/lava/proc/lif/test_models.py
ackurth-nc: Addressed in 8fa7f87.

tests/lava/proc/lif/test_models.py
ackurth-nc: Addressed in 8fa7f87.

src/lava/proc/lif/models.py
ackurth-nc: I agree that this is unintuitive.
I followed your suggestion of introducing a local copy with promoted ```dtype```, see 0b2c64b.

mathisrichter (approved):

PhilippPlank (approved):

Issue #398, opened 2022-10-06
Add floating2fixed point converter


ackurth-nc:
Objective of issue:
Add  tool to perform floating- to fixed-point conversion as suggested [here](https://github.com/lava-nc/lava/blob/main/src/lava/utils/float2fixed.py).

**I'm submitting a ...** 

- [ ] bug report
- [x] feature request
- [ ] documentation request

mathisrichter:
@awintel What is the status of this issue/PR?

awintel:
Last year, @ackurth-nc told me he would finish the PR and bring it to a conclusion. That's all I know at the moment.

PR #399, opened 2022-10-06
Float2Fixed converter

ackurth-nc:
Issue Number: #398 

Objective of pull request: Add tool performing automated floating- to fixed-point conversion.

## Pull request checklist

- [] [Docs](https://github.com/lava-nc/docs) reviewed and added / updated if needed (for bug fixes / features)

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] Documentation changes

- [ ] Yes
- [x] No


src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
-5. Initialize fixed-point ProcessModels
-
-ProcessModel extraction might be a shared function with compiler.
-
-This will likely lead to some performance degradation of the application so
-perhaps some iterative optimization may have to be included within this
-parameter conversion process. Reuse other libraries for this purpose where
-possible.
- """
+This module containx a tool to perform automated float- to fixed-point
```
PhilippPlank, line 5: contains*

Move this description into the class? The module seems to be the Float2FixedConverter class.

tests/lava/magma/core/model/py/test_type.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def test_lavapytype_constructor(self):
+        """Check if LavaPyType is correctly instantiated"""
+        lava_py_type = LavaPyType(cls=None, d_type=None)
+
+        # Check if correct instance is created.
+        self.assertIsInstance(lava_py_type, LavaPyType)
+
+    def test_validate_precision_input_warning(self):
+
```
PhilippPlank, line 20: Some docstring briefly explaining what the unittest checks would be good. (Also for the other tests)

tests/lava/tutorials/test_tutorials.py
```diff
@@ -216,7 +216,7 @@ def test_end_to_end_01_mnist(self):
             "tutorial01_mnist_digit_classification.ipynb", e2e_tutorial=True
         )
 
-    @unittest.skipIf(system_name != "linux", "Tests work on linux")
+    @unittest.skip('Running the notebook takes too long')
```
PhilippPlank, line 219: We might need some special tests for tutorials then, otherwise broken tutorials will end up in releases. @mgkwill any ideas?

src/lava/magma/core/model/py/type.py
```diff
@@ -4,6 +4,12 @@
 import typing as ty
 from dataclasses import dataclass
 from lava.magma.core.model.py.ports import PyInPort, PyOutPort, PyRefPort
+import numpy as np
+import warnings
+
+# Members of LavaPyType needed for float- to fixed-point conversion.
+CONST_CONV_VARS = ('num_bits_exp', 'domain', 'constant', 'scale_domain',
```
PhilippPlank, line 11: Conv might be mistaken with convolution. Also, does this constant need to be globally defined? Seems to be used only in conversion_data(..).

PhilippPlank (commented):
Looks good overall. I have not checked the conversion in great detail. Left some remarks.

src/lava/magma/core/model/py/type.py
```diff
@@ -12,4 +18,104 @@ class LavaPyType:
...
+            If precision is not of type str or None
+        ValueError
+            If precision passed in wrong format
+        """
+
+        if precision is None:
+            warnings.warn("'precision' is None: This might cause an error in"
+                          + " the Float2FixedPoint conversion")
+        elif type(precision) is not str:
+            raise TypeError("'precision' must be of type str or None")
```
phstratmann, line 59: When raising an error it is best to include the actual value of the variable for debugging. Thus, I would here add the precision variable. Also below in the ValueErrors. Something of the form

```
raise TypeError(f"'precision' must be of type str or None, but'"
    f"{is of type type(precision)}'")
```

src/lava/magma/core/model/py/type.py
```diff
@@ -12,4 +18,104 @@ class LavaPyType:
...
+                raise ValueError("Second string literal of precision (number"
+                                 + " of  bits) must be interpretable as int")
+            try:
+                int(split_precision[2])
+            except ValueError:
+                raise ValueError("Thirds string literal of precision (implicit"
+                                 + " shift) must be interpretable as int")
+
+    def _validate_exp_data(self):
+        """Validates if data regarding in a split of the variable in mantissa
```
phstratmann, line 81: In general, there are still a few typos or grammatical errors in docstrings. Maybe worth to have a final look at it, if you find time.

src/lava/magma/core/model/py/type.py
```diff
@@ -12,4 +18,104 @@ class LavaPyType:
         type, ty.Type[PyInPort], ty.Type[PyOutPort], ty.Type[PyRefPort]
     ]
     d_type: type
-    precision: int = None  # If None, infinite precision is assumed
+    precision: str = None  # If None, infinite precision is assumed.
```
phstratmann, line 21: Why not `math.inf` for infinite precision?

src/lava/proc/dense/models.py
```diff
@@ -59,13 +59,18 @@ class PyDenseModelBitAcc(PyLoihiProcessModel):
     it mimics Loihi behavior bit-by-bit.
     """
 
-    s_in: PyInPort = LavaPyType(PyInPort.VEC_DENSE, bool, precision=1)
-    a_out: PyOutPort = LavaPyType(PyOutPort.VEC_DENSE, np.int32, precision=16)
-    a_buff: np.ndarray = LavaPyType(np.ndarray, np.int32, precision=16)
+    s_in: PyInPort = LavaPyType(PyInPort.VEC_DENSE, bool, precision='u:1:0')
```
phstratmann, line 62: So great to see that this precision feature is finally implemented! :-)

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
 # Copyright (C) 2021-22 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
 # See: https://spdx.org/licenses/
-
 """
-This module will contain a tool to perform automated float- to fixed-point
```
phstratmann, line 6: will? It does not yet?

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+from lava.magma.core.process.ports.ports import AbstractSrcPort
+from lava.proc.monitor.process import Monitor
+from lava.magma.core.process.ports.ports import OutPort
+
+import numpy as np
+import inspect
+
+
+class Float2FixedConverter:
+
```
phstratmann, line 24: Docstring missing. Probably a slightly extended version of lines 5-6

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+    def __init__(self):
+        self.floating_pt_rcfg = None
```
phstratmann, line 26: write run_cfg. Took me a few seconds to figure out what this abbreviation means, and it's not worth that hassle just to save three characters.

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+        self.procs = None
+        self.hierarchical_procs = None
+        self.run_proc = None
+        self.var_ports = None
+        self.fixed_pt_proc_models = None
+        self.num_steps = None
+        self.quantiles = None
+        self.scale_domains = set()
+
+    def set_run_cfg(self, floating_pt_rcfg, fixed_pt_rcfg) -> None:
```
phstratmann, line 37: I didn't see where set_run_cfg is being used.
But maybe it's easier/cleaner to simply write setter methods for floating_pt_rcfg und fixed_pt_rcfg?
https://www.geeksforgeeks.org/getter-and-setter-in-python/

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+
+        self.floating_pt_rcfg = floating_pt_rcfg
+        self.fixed_pt_rcfg = fixed_pt_rcfg
+
+    def convert(self, proc, num_steps=100, quantiles=[0, 1],
+                find_connected_procs=True):
+        """Convert Processes passed from floating-point to fixed-point model.
+        The floating- and fixed-point ProcessModels to be used must be
+        specificed via setting the run configurations.
+        After conversion has been executed, the converter parameter are stored
```
phstratmann, line 62: parameter are -> singular or plural? ;-)

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+               converted, instantiate new Processes copying the structure of
+               the passed model without hierarchical Processes. In the
+               following, the conversion will be performed on the new model.
+               Redo step 2 with the new model and continue with 4.
+            4. Finding and instantiating the fixed-point ProcessModels.
+            5. Retrieving the information needed for conversion from the
+               fixed-point ProcessModels chosen via the provided fixed-point
+               run configuration.
+            5. Creating monitors for recording the activation statistics of the
+            6  dynamical variables of Processes. With this, the domain of the
```
phstratmann, line 79: of the "6" dynamical variables?
Why are there two fifths ("5.") bullet points? :-D

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+               variable can be determined to constrain the floating- to
+               fixed-point conversion.
+            7. Execution of the floating-point model chosen via run
+               configuration. If non-default quantile information is passed the
```
phstratmann, line 83: What's a non-default quantile information?

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+            variables which is needed for the conversion but also will lead to
+            a longer run time
+        quantiles : list
+            List of two numbers both in the interval [0,1] determining the
+            p-quantile of the domain of a monitored variable to set largest and
+            smallest value of the dynamic rep_range of that variable. The
+            dynamic range will be used to determine the optimal floating- to
+            fixed-point conversion. The first entry must be smaller than the
+            second. 0 resp. 1 correspond to the minimum resp. maximum of the
+            domain.
```
phstratmann, line 108: "0 resp. 1 correspond to the minimum resp. maximum of the domain."
That sentence seems redundant to me.

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+        if not issubclass(type(floating_pt_rcfg), RunConfig):
+            raise TypeError("'floating_pt_rcfg' must be RunConfig")
+
+        if not issubclass(type(fixed_pt_rcfg), RunConfig):
+            raise TypeError("'fixed_pt_rcfg' must be RunConfig")
+
+        self.floating_pt_rcfg = floating_pt_rcfg
+        self.fixed_pt_rcfg = fixed_pt_rcfg
+
+    def convert(self, proc, num_steps=100, quantiles=[0, 1],
```
phstratmann, line 57: maybe better to define quantiles  using a tuple, rather than a list?
This tuple will never be changed, and thus a list is unnecessary overhead.

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+        # Infer Variables and Ports of Processes.
+        self.var_ports = self._get_var_ports()
+        # If hierarchical Processes are involved, set up copy of model where
+        # Subprocesses are made explicit.
+        if self.hierarchical_procs:
+            self._explicate_hierarchical_procs()
+        else:
+            pass
+        # Get fixed-point ProcessModels of Processes.
+        self.fixed_pt_proc_models = self._get_fixed_pt_proc_models()
```
phstratmann, line 124: What's the difference between line 124 and 126?

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+        self._set_procs(proc=proc, find_connected_procs=find_connected_procs)
+        # Infer Variables and Ports of Processes.
+        self.var_ports = self._get_var_ports()
+        # If hierarchical Processes are involved, set up copy of model where
+        # Subprocesses are made explicit.
+        if self.hierarchical_procs:
+            self._explicate_hierarchical_procs()
+        else:
+            pass
+        # Get fixed-point ProcessModels of Processes.
```
phstratmann, line 123: I like the way you handle the following code. Code is hidden in separate private methods, and thus it is very clear in the following what you are doing. Great!

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+        """
+        if issubclass(type(proc), AbstractProcess):
+            if find_connected_procs:
+                proc_list = find_processes(proc)
+            else:
+                proc_list = [proc]
+        elif isinstance(proc, list):
+            for p in proc:
+                if not issubclass(type(p), AbstractProcess):
+                    raise TypeError("list 'proc' must contains Processes")
```
phstratmann, line 162: "contain"

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+                proc_list = proc
+        else:
+            raise TypeError("'proc' must be Process of list of Processes")
```
phstratmann, line 165: "or"
Also here, return proc or (here better) its type

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+
+        Parameters
+        ----------
+
+        proc : AbstracProcess or List[AbstractProcess]
+            Processes to be converted
+        find_connected_procs : bool
+            Find all Processes connected to passed Process
+        """
+        if issubclass(type(proc), AbstractProcess):
```
phstratmann, line 154: According to the docstring, it sounds if if the user should pass...
... either a process, and the method finds all linked processes
... or a list of processes.
Thus, the bool find_connected_procs can be auto-filled, or isn't that the case?

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+        # Get ProcessModel classes of Processes with fixed-point RunConfig.
+        p_model_cls = ProcGroupDiGraphs._map_proc_to_model(
+            proc_list,
+            self.floating_pt_rcfg)
+
+        # Instantiate fixed-point ProcessModel of Processes and put them into
+        # dictionary. Moreover we update the self.procs from a new proc list
+        # containing Subprocesses but not their Hierarchical Processes anymore.
+        proc_list_w_sub = []
+        for p, model_cls in p_model_cls.items():
```
phstratmann, line 176: What do you need model_cls for? If not, just use p_model_cls.keys()

Side note: it should be sufficient and substantially faster to write
`proc_list_w_sub = p_model_cls.keys()`

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+        # hierarchical Processes.
+        self.procs = dict([(p.id, p) for p in proc_list_w_sub])
+        self.hierarchical_procs = dict([(p.id, p) for p in
+                                        hierarchical_procs_set])
+
+        self.run_proc = proc_list[0]
+
+    def _explicate_hierarchical_procs(self):
+        """Given a Model with hierarchical Processes, set up an equivalent
+        model with only first-order Processes, i.e. not hierarchical Processes.
```
phstratmann, line 192: PReviously, we called this leaf processes

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+        This is needed for the subsequent monitoring of Process Variables and
+        Conversion. The newly created Processes are instantiated with the same
+        intial values as all (Sub)Processes in the original model."""
+
+        # Create dictionary for storing the conenction date.
+        conn_data = {}
+
+        for p_id, p_var_ports in self.var_ports.items():
+            # Increase depth of dictionary to collect connection data.
+            conn_data[p_id] = {}
```
phstratmann, line 202: Above, you mention that "The newly created Processes are instantiated with the same intial values as all (Sub)Processes in the original model."
Thus, will line 202 be executed with the same p_id several times? If so, entries of the dict will be overwritten.

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+
+            p = self.procs[p_id]
+
+            for out in p_var_ports['SourcePort']:
```
phstratmann, line 206: Maybe faster to create this using dict comprehension?

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+                        t_port_conv_data = t_port_py_type.conversion_data()
+                    except (TypeError, ValueError):
+                        raise ValueError(f'Error in {t_port} of Proc {p_id}')
+                    t_port_impl_shift = t_port_conv_data['implicit_shift']
+                    # If there has been not implicit shift set so far set one.
+                    if not p_impl_shift:
+                        p_impl_shift = t_port_impl_shift
+
+                    # Check for consistency of implicit shifts due to Ports
+                    elif not p_impl_shift == t_port_impl_shift:
```
phstratmann, line 402: Where could this inconsistent shift come from? It looks like there is a redundant way for the user to set this. Is there any way we may prevent this in general?

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+        -------
+        monitors : dict
+            Monitoring infrastructure for Float2Fixed converter in nested dict
+            {ProcID:
+                {VarName:
+                     monitor}}
+        """
+        monitors = {}
+
+        for p_id, conv_data_p_id in self.conv_data.items():
```
phstratmann, line 458: Here, it should be possible to replace nested for-loops by a dict comprehension.

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+        """Run Processes passed to converter with given number of steps, get
+        and store data of monitored Variables in conversion data dictionary.
+        The needed 'run_config' is taken from the passed 'floating_pt_rcfg'."""
+        rcfg = self.floating_pt_rcfg
+        run_cond = RunSteps(num_steps=self.num_steps)
+
+        # Run Processes from designated Process.
+        self.run_proc.run(condition=run_cond, run_cfg=rcfg)
+        # Get monitored variables.
+        for p_id, monitor_dict in self.monitors.items():
```
phstratmann, line 487: Dict comprehension?

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+                   {VarName:
+                        {'precision': _,
+                         'domain': _,
+                         'constant: _,
+                         'num_bits_exp':_}}}}
+        """
+
+        # Initialize scale domain dictionary.
+        scale_domains = {}
+        for scale_domain, p_id in self.scale_domains:
```
phstratmann, line 519: dict comprehension

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+                for var, scale_data in scale_data_p_id.items():
+
+                    bits = scale_data['precision_bits']
+                    impl_shift = scale_data['implicit_shift']
+                    num_bits_exp = scale_data['num_bits_exp']
+
+                    # Get range representable given bit constraints and
+                    # signedness.
+                    if scale_data['signedness'] == 's':
+                        rep_range = np.array([-1 * 2 ** (bits - 1),
```
phstratmann, line 572: We should decide how to represent ranges. Above, you used a list. Here an array. I think the best way would be a tuple, or what do you think?

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+                                              2 ** (bits - 1) - 1])
+                    elif scale_data['signedness'] == 'u':
+                        rep_range = np.array([0, 2 ** bits - 1])
+
+                    # Apply implicit shift.
+                    rep_range = np.left_shift(rep_range, impl_shift)
+
+                    # Apply maximal shift due to exponent.
+                    rep_range = np.left_shift(rep_range, 2 ** num_bits_exp - 1)
```
phstratmann, line 581: Given that you're a mathematician, a general question:
For quite a while, I was confused in our team that _exp typically refers to powers of two. Intuitively, it refers to an exponential function. Can you think of any other way how to abbreviate this?

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+
+                    # Get range of Variables in floating point domain.
+                    domain = scale_data['domain']
+
+                    # If domain of Variable has been recorded apply quantile.
+                    if var in self.monitors[p_id].keys():
+                        lower_val = np.quantile(domain, self.quantiles[0])
+                        higher_val = np.quantile(domain, self.quantiles[1])
+                        dyn_range = np.array([lower_val, higher_val])
```
phstratmann, line 590: Also here, how to represent ranges? Array, list, tuple? I would argue for tuples.

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
+                    # Calculate slopes for linear mapping of var.
+                    # If an entry in rep_range or dyn_range is zero, inf will
+                    # be filled in.
+                    not_zero_cond = (rep_range != 0) * (dyn_range != 0)
+                    slopes_var = np.divide(rep_range, dyn_range,
+                                           out=np.inf * np.ones_like(dyn_range),
+                                           where=not_zero_cond)
+
+                    # Take absolute value. Only positive slopes are allowed.
+                    slopes_var = np.absolute(slopes_var)
```
phstratmann, line 603: Is there a way how the slope may be negative? If so, there is something wrong. In that case, the absolute value of the slope would be a random number, and I would suggest that we throw an error instead. Or what do you think?

src/lava/utils/float2fixed.py
```diff
@@ -1,26 +1,719 @@
...
-ProcessModel extraction might be a shared function with compiler.
-
-This will likely lead to some performance degradation of the application so
-perhaps some iterative optimization may have to be included within this
-parameter conversion process. Reuse other libraries for this purpose where
-possible.
- """
+This module containx a tool to perform automated float- to fixed-point
+conversion of Lava Processes.
+"""
```
phstratmann, line 7: A general remark:
As a user, I may often not know what the ideal quantiles are for my variables. That strongly depends on the distribution of the recorded data. Maybe it is worth to add a method that - after recording - plots the recorded distribution and allows the user to visually choose a quantile.
I guess such a util is out of scope for your internship. If so, maybe add a github issue instead.

tests/lava/tutorials/test_tutorials.py
phstratmann: In NxSDK, long tests are only executed in dedicated runs during the night. Maybe that would be an option?

tests/lava/test_utils/test_float2fixed.py
```diff
@@ -0,0 +1,1553 @@
...
+                                  precision="u:16:3")
+    outport: PyOutPort = LavaPyType(PyOutPort.VEC_DENSE, np.int16,
+                                    precision="u:16:0")
+
+
+@implements(proc=Proc, protocol=LoihiProtocol)
+@requires(CPU)
+@tag('fixed_pt_pytype_var_err')
+class ProcPyProcModelVarErr(PyLoihiProcessModel):
+    """A minimal fixed-point PyProcModel implementing Proc."""
```
phstratmann, line 104: The docstrings should mention inhowfar the different PyProcModels differ mutually.

tests/lava/test_utils/test_float2fixed.py
```diff
@@ -0,0 +1,1553 @@
...
+@requires(CPU)
+@tag('fixed_pt')
+class ProcVarPyProcModelFixed(PyLoihiProcessModel):
+    """A minimal fixed-point PyProcModel implementing ProcVar."""
+
+    v: np.ndarray = LavaPyType(np.ndarray, np.int32, precision="s:16:3",
+                               constant=False, scale_domain=0)
+
+
+class Float2Fixed(unittest.TestCase):
```
phstratmann, line 370: The class should be termed "TestFloat2Fixed"

src/lava/utils/float2fixed.py
ackurth-nc: Addressed in e377d035f0901c0c8201ce71e8372d2724e0f03f.

tests/lava/magma/core/model/py/test_type.py
ackurth-nc: Addressed in 8092d313c76b4e45680cfc4aee491210ae60321a.

src/lava/magma/core/model/py/type.py
ackurth-nc: Addressed in 0117a669a3f2c37e943eefb164a56100d4267f21.

src/lava/magma/core/model/py/type.py
ackurth-nc: Addressed in 96d55e369307c301015c073d3ccec2049e89b006.

src/lava/magma/core/model/py/type.py
ackurth-nc: Addressed by introducing precision class, see 96d55e369307c301015c073d3ccec2049e89b006.

src/lava/magma/core/model/py/type.py
ackurth-nc: Addressed in ffc5674879659181d40cb0925d5e82be755d2a1f.

src/lava/utils/float2fixed.py
ackurth-nc: Addressed in e377d035f0901c0c8201ce71e8372d2724e0f03f.

src/lava/utils/float2fixed.py
ackurth-nc: Good point, addressed in 4a8c69297ec83114efe9491991d0328736556207.

src/lava/utils/float2fixed.py
ackurth-nc: Addressed in 58fd18a6c5ec776b81b991a6260f4edcc2d2ca2d.

src/lava/utils/float2fixed.py
ackurth-nc: Addressed in 58fd18a6c5ec776b81b991a6260f4edcc2d2ca2d.

src/lava/utils/float2fixed.py
ackurth-nc: Addressed in 58fd18a6c5ec776b81b991a6260f4edcc2d2ca2d.

src/lava/utils/float2fixed.py
ackurth-nc: Addressed in 58fd18a6c5ec776b81b991a6260f4edcc2d2ca2d.

src/lava/utils/float2fixed.py
ackurth-nc: Nice point. Addressed in 58fd18a6c5ec776b81b991a6260f4edcc2d2ca2d.

src/lava/utils/float2fixed.py
ackurth-nc: Nice, should be addressed in 58fd18a6c5ec776b81b991a6260f4edcc2d2ca2d.

src/lava/utils/float2fixed.py
ackurth-nc: Addressed in 58fd18a6c5ec776b81b991a6260f4edcc2d2ca2d.

src/lava/utils/float2fixed.py
ackurth-nc: Renamed in 58fd18a6c5ec776b81b991a6260f4edcc2d2ca2d.

tests/lava/test_utils/test_float2fixed.py
ackurth-nc: Addressed in 8dc7625d98f36094ce348d3af28f3cc5cbd3b089.

src/lava/utils/float2fixed.py
ackurth-nc: True, addressed in feb0b3aab0d2bc85d74e6380035de06d84b4f615.

src/lava/utils/float2fixed.py
ackurth-nc: I don't think there would be something wrong.
Imagine the following situation:  
For a variable `var` we use a signed variable with `7` bits, implying smallest and highest values are `-2**6` and `2**6-1*`.  
Given a specific parameter choice, `var` only attains positive values, say the smallest is `0.1` and the largest is `1.`  The slopes are calculated as  `-2**6 / 0.1` and `(2**6 -1) / 1`.    
The first now is negative, but this is now problem. Since we want, however, the smallest positive slope we calculate the absolute value of the slopes and then take the minimum.  
Does this address your question?

src/lava/utils/float2fixed.py
ackurth-nc: In principal I agree.  
But I don't really know a better name, something like "exponent to basis 2" maybe, but this does not admit a nice abbreviation.

But if you have a better idea than `num_bits_exp`  I readily change it.

src/lava/utils/float2fixed.py
ackurth-nc: Thanks. :)

src/lava/utils/float2fixed.py
ackurth-nc: Done.

src/lava/utils/float2fixed.py
ackurth-nc: I thought about that, but I think it would significantly reduce the readability of the code.  
But since I am not the best person the judge the readability I might be wrong.  
How do you think about the comprehension in wrt the readability?

src/lava/utils/float2fixed.py
ackurth-nc: Say you write you custom adapted neuron model - let's call it PLIF - which has a different implicit shift then the classical LIF implemented in Lava.  
Now you instantiate a `plif` and `lif` process and connect it to the same `dense` process.  
In this case, we would have inconsistent implicit shifts.  
I think in general this is not preventable.

src/lava/utils/float2fixed.py
ackurth-nc: Same point as above - readability is I think more important.  
But I am open to your opinion whether or not to change it.

src/lava/utils/float2fixed.py
ackurth-nc: See above.

src/lava/magma/core/model/py/type.py
```diff
@@ -12,4 +18,104 @@ class LavaPyType:
...
+        Raises
+        ------
+        ValueError
+            If exponent data is not complete
+        """
+
+        if self.num_bits_exp and not self.exp_var:
+            raise ValueError("Provided number of bits for exponent but no"
+                             + " name for exponent variable.")
+        if self.exp_var and not self.num_bits_exp:
```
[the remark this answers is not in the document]
ackurth-nc: True, corrected in d388bc20ddc0cd057ce1f7aec71bfb49d4613bc2.

src/lava/utils/float2fixed.py
ackurth-nc: `set_run_cfg` is used be the user before calling the `converter.convert`, see [here, cell 11](https://github.com/ackurth-nc/lava/blob/dev/f2f_conv/tutorials/end_to_end/tutorial02_oxford_input_transformation.ipynb).

ackurth-nc:
Right now the CI does no go through.
If I understand the error correctly, the problem is not due to a change introduced by me.
Could someone confirm or point me to where the problem is - I am unable to identify it.

PhilippPlank:
> Right now the CI does no go through. If I understand the error correctly, the problem is not due to a change introduced by me. Could someone confirm or point me to where the problem is - I am unable to identify it.
> 
> Also the removal from the reviewers who did not give a comment yet was unintentional. Apparently GitHub does that automatically when requesting new reviews.

I fixed the issue. Automatic testing of jupyter tutorials can currently only be done for Linux OS, since the converting pipeline is build that way. It failed for Windows and MacOS.

src/lava/utils/float2fixed.py
ackurth-nc: No, because the user also might want to pass a list of processes they want to convert and not detect the other ones.

src/lava/utils/float2fixed.py
ackurth-nc: Addressed in e29f938e480575f0dd5e535280d0ce4271f3a1c8

mgkwill:
Merging this into lava-nc branch f2f_conv so this can be added to without access to the fork.

PR #418, opened 2022-10-14
Compiler graph: Enable model selection, fixing warning in QUBO tutorial

phstratmann:
Issue Number: 417 (& 56 in lava-optimization)

Objective of pull request:
As the compiler graph scans through all modules in the path, it checks if either of the encountered class is a ProcModel.
If a file imports a module, like in solver.py of lava-optimization,
from lava.lib.optimization.solvers.generic.sub_process_models import StochasticIntegrateAndFireModel
the crawler will try to import that module as well. This leads to a warning, as the compiler graph incorrectly assumes that the found StochasticIntegrateAndFireModel is in solver.py.
The correct behavior should be that the crawler goes through all modules, and only selects Models implemented in the specific module, not those that are only imported.

The important change is in compiler_graphs.py.
Unfortunately, this change broke four unit tests, three in test_monitors, one in test_tutorial. This was due to the fact that the new crawler chooses the bitacc LifProcModel by default. Accordingly, I now specifically select the floating_pt model where tests would break otherwise.

## Pull request checklist

## Pull request type

Please check your PR type:
- [x] Bugfix


PhilippPlank (approved):

GaboFGuerra (approved):
Thanks for fixing this!

PR #443, opened 2022-10-25
Add composite action CI Setup rename ci-build.yml to ci.yml

mgkwill:


mgkwill:
@PhilippPlank @mathisrichter  FYI - I've moved to a composite action in https://github.com/lava-nc/ci-setup-composite-action since there is so much duplication in our jobs.

PR #467, opened 2022-11-08
Fix test_exception_handling.py

shaline-koh:
With the implementation of C++ backend for message passing, `test_exception_handling.py` no longer passes as the exception is not passed to the parent process.

This PR implements a workaround by creating a shared memory block which increments a counter when `run_spk` is called and `time_step > 1`. When the process terminates, the test case will assert that the counter has been incremented.


seowwj:
Unit test CI is currently failing at `test_file_descriptor.py` with `SegmentationFault`

tests/lava/magma/runtime/test_exception_handling.py
```diff
@@ -93,19 +123,17 @@ def test_one_pm(self):
...
+        proc.run(condition=run_steps, run_cfg=run_cfg)
 
-        exception = context.exception
-        self.assertEqual(RuntimeError, type(exception))
-        # 1 exception in the ProcessModel expected
-        self.assertTrue('1 Exception(s) occurred' in str(exception))
+        # Check that the error count has increased to 1
+        existing_shm = shared_memory.SharedMemory(name='error_block')
+        res = np.ndarray((1,), buffer=existing_shm.buf)
+        self.assertEqual(res[0], 1)
```
killight98, line 131: - could mv `existing_shm.close()` before assert func?
- add proc.stop()

tests/lava/magma/runtime/test_exception_handling.py
```diff
@@ -121,19 +149,17 @@ def test_two_pm(self):
...
+        sender.run(condition=run_steps, run_cfg=run_cfg)
 
-        exception = context.exception
-        self.assertEqual(RuntimeError, type(exception))
-        # 2 Exceptions in the ProcessModels expected
-        self.assertTrue('2 Exception(s) occurred' in str(exception))
+        # Check that the error count has increased to 2
+        existing_shm = shared_memory.SharedMemory(name=self.shm_name)
+        res = np.ndarray((1,), buffer=existing_shm.buf)
+        self.assertEqual(res[0], 2)
```
killight98, line 157: same with test_one_pm

tests/lava/magma/runtime/test_exception_handling.py
```diff
@@ -150,14 +176,14 @@ def test_three_pm(self):
...
```
killight98, line 184: same with test_one_pm

killight98 (changes requested):

tests/lava/magma/runtime/test_exception_handling.py
```diff
@@ -93,19 +123,18 @@ def test_one_pm(self):
...
-            proc.run(condition=run_steps, run_cfg=run_cfg)
+        proc.run(condition=run_steps, run_cfg=run_cfg)
 
-        exception = context.exception
-        self.assertEqual(RuntimeError, type(exception))
-        # 1 exception in the ProcessModel expected
-        self.assertTrue('1 Exception(s) occurred' in str(exception))
+        # Check that the error count has increased to 1
+        existing_shm = shared_memory.SharedMemory(name='error_block')
+        res = np.ndarray((1,), buffer=existing_shm.buf)
```
killight98, line 130: add np.copy as following:
```py
res = np.copy(np.ndarray((1,), buffer=existing_shm.buf))
```
others are same.

Issue #599, opened 2023-02-01
Make Runtime a Python Context Manager


Gavinator98:
### User story
As a user, I want the Lava runtime to utilize [Python context managers](https://docs.python.org/3/reference/datamodel.html#context-managers) so it is easier to automatically call stop().

The proposed pattern of a runtime would be:
```python
with process_to_run:
    process_to_run.run(condition=RunSteps(num_steps=1000), run_cfg=run_config)
```

If there is interest in this design, I would be happy to implement it.

### Conditions of satisfaction
- A Python context manager can be used to automatically stop runtime.
- [other use cases?]
- Feature is fully unit tested for every use case.
- Public interface of the feature is documented with docstrings, where appropriate.
- Lava coding conventions are met.

mathisrichter:
Hi @Gavinator98 

That is a great idea, thanks for proposing it and volunteering to work on it. We are very happy to support you in developing this nice feature.

Feel free to implement it and open a PR, linking back to this issue. We will review the PR and possibly suggest changes.

Before you start, please have a look at our [Lava developer guide](https://lava-nc.org/developer_guide.html), in particular the section on [Coding Conventions](https://lava-nc.org/developer_guide.html#coding-conventions). The first section of the guide is somewhat outdated but the coding conventions are still up-to-date.

For the PR, there need to be unit tests for all use cases of this feature as well as documentation in terms of docstrings, where appropriate.

We can use the comment section here to communicate if there are questions.
Please do not hesitate to reach out for clarification or help!

Mathis

PR #605, opened 2023-02-04
Update runtime and process to be "with" context managers.

Gavinator98:
Issue Number: #599 

Objective of pull request: Update runtime and processes to be  [Python "with" context managers](https://docs.python.org/3/reference/datamodel.html#context-managers) so it is easier to automatically call stop(). 

## Pull request checklist

- [X] [Issue](https://github.com/lava-nc/lava/issues) created that explains the change and why it's needed
- [X] Tests are part of the PR (for bug fixes / features)
- [X] PR conforms to [Coding Conventions](https://lava-nc.org/developer_guide.html#coding-conventions)
- [X] [PR applys BSD 3-clause or LGPL2.1+ Licenses](https://lava-nc.org/developer_guide.html#add-a-license) to all code files
- [X] Lint (`flakeheaven lint src/lava tests/`) and (`bandit -r src/lava/.`) pass locally
- [X] Build tests (`pytest`) passes locally

## Pull request type

 

 

Please check your PR type:

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] Documentation changes

- Runtime and processes do not support "with" statements.

## What is the new behavior?

- Runtime and processes both support "with" statements.
- When entering a with block the runtime is initialized
- When exiting a with block the runtime is stopped.

### Example:
```
with process_to_run:
    process_to_run.run(condition=RunSteps(num_steps=1000), run_cfg=run_config)
```

- [ ] Yes
- [X] No

### DIFF
--- a/src/lava/magma/core/process/process.py
+++ b/src/lava/magma/core/process/process.py
@@ -7,7 +7,6 @@ import logging
 import typing as ty
 from _collections import OrderedDict
 from dataclasses import dataclass
-from contextlib import AbstractContextManager
 from lava.magma.compiler.executable import Executable
 
 from lava.magma.core.process.interfaces import \
@@ -36,7 +35,7 @@ class ProcessPostInitCaller(type):
         return obj
 
 
-class AbstractProcess(AbstractContextManager, metaclass=ProcessPostInitCaller):
+class AbstractProcess(metaclass=ProcessPostInitCaller):
     """The notion of a Process is inspired by the Communicating Sequential
     Process paradigm for distributed, parallel, and asynchronous programming.
 
@@ -221,6 +220,10 @@ class AbstractProcess(AbstractContextManager, metaclass=ProcessPostInitCaller):
         free compute resources.
         """
         self.stop()
+
+    def __enter__(self):
+        """Required for "with" block."""
+        pass
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         """Stop the runtime when exiting "with" block of a context manager."""
--- a/src/lava/magma/core/process/process.py
+++ b/src/lava/magma/core/process/process.py
@@ -226,9 +226,7 @@ class AbstractProcess(metaclass=ProcessPostInitCaller):
         pass
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        """
-        Stop the runtime when exiting "with" block.
-        """
+        """Stop the runtime when exiting "with" block of a context manager."""
         self.stop()
 
     def _post_init(self):
--- a/src/lava/magma/core/process/process.py
+++ b/src/lava/magma/core/process/process.py
@@ -221,6 +221,16 @@ class AbstractProcess(metaclass=ProcessPostInitCaller):
         """
         self.stop()
 
+    def __enter__(self):
+        """Required for "with" block."""
+        pass
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """
+        Stop the runtime when exiting "with" block.
+        """
+        self.stop()
+
     def _post_init(self):
         """Called after __init__() method of any sub class via
         ProcessMetaClass to finalize initialization leading to following
--- a/src/lava/magma/core/process/process.py
+++ b/src/lava/magma/core/process/process.py
@@ -222,7 +222,7 @@ class AbstractProcess(metaclass=ProcessPostInitCaller):
         self.stop()
 
     def __enter__(self):
-        """Required for "with" block."""
+        """Executed when Process enters a "with" block of a context manager."""
         pass
 
     def __exit__(self, exc_type, exc_val, exc_tb):
--- a/src/lava/magma/core/process/process.py
+++ b/src/lava/magma/core/process/process.py
@@ -7,6 +7,7 @@ import logging
 import typing as ty
 from _collections import OrderedDict
 from dataclasses import dataclass
+from contextlib import AbstractContextManager
 from lava.magma.compiler.executable import Executable
 
 from lava.magma.core.process.interfaces import \
@@ -35,7 +36,7 @@ class ProcessPostInitCaller(type):
         return obj
 
 
-class AbstractProcess(metaclass=ProcessPostInitCaller):
+class AbstractProcess(AbstractContextManager, metaclass=ProcessPostInitCaller):
     """The notion of a Process is inspired by the Communicating Sequential
     Process paradigm for distributed, parallel, and asynchronous programming.
 
@@ -220,10 +221,6 @@ class AbstractProcess(metaclass=ProcessPostInitCaller):
         free compute resources.
         """
         self.stop()
-
-    def __enter__(self):
-        """Required for "with" block."""
-        pass
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         """Stop the runtime when exiting "with" block of a context manager."""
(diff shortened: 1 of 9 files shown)


weidel-p (approved):
That's a clean and easy implementation and include tests.

If you want, you could demonstrate this feature in the tutorial about execution:
https://github.com/lava-nc/lava/blob/main/tutorials/in_depth/tutorial04_execution.ipynb

PhilippPlank (approved):

gkarray (approved):

src/lava/magma/runtime/runtime.py
```diff
@@ -135,6 +135,14 @@ def __del__(self):
         if self._is_started:
             self.stop()
 
+    def __enter__(self):
+        """Initialize the runtime on entering a "with" block"""
```
mathisrichter, line 139: Missing full-stop at the end of the sentence.

src/lava/magma/core/process/process.py
```diff
@@ -221,6 +221,16 @@ def __del__(self):
         """
         self.stop()
 
+    def __enter__(self):
+        """Required for "with" block."""
```
mathisrichter, line 225: It could be clearer here to explicitly refer to context managers.
```python
"""Executed when Process enters a "with" block of a context manager."""
```

src/lava/magma/core/process/process.py
```diff
@@ -221,6 +221,16 @@ def __del__(self):
...
+        pass
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """
+        Stop the runtime when exiting "with" block.
+        """
```
mathisrichter, line 231:
```python
"""Stop the runtime when exiting "with" block of a context manager."""
```

tests/lava/magma/runtime/test_context_manager.py
```diff
@@ -0,0 +1,114 @@
+import unittest
```
mathisrichter, line 1: This is missing a license header. Please add this at the top of the file and leave a blank line to the first import statement.
```python
# Copyright (C) 2021-22 Intel Corporation
# SPDX-License-Identifier: BSD-3-Clause
# See: https://spdx.org/licenses/
```

tests/lava/magma/runtime/test_context_manager.py
```diff
@@ -0,0 +1,114 @@
...
+        self.u = Var(shape=shape, init=0)
+        self.v = Var(shape=shape, init=0)
+
+
+@implements(proc=SimpleProcess, protocol=LoihiProtocol)
+@requires(CPU)
+class SimpleProcessModel(PyLoihiProcessModel):
+    """
+    Defines a SimpleProcessModel
+    """
```
mathisrichter, line 32: This docstring can be removed as it does not add additional information.

tests/lava/magma/runtime/test_context_manager.py
```diff
@@ -0,0 +1,114 @@
...
+    u = LavaPyType(int, int)
+    v = LavaPyType(int, int)
+
+
+class SimpleRunConfig(RunConfig):
+    """
+    Defines a simple run config
+    """
```
mathisrichter, line 40: This docstring can be removed as it does not add additional information.

tests/lava/magma/runtime/test_context_manager.py
```diff
@@ -0,0 +1,114 @@
...
+        Ensures process/runtime is stopped if context manager fails to
+        """
+        self.stoppable.stop()
+
+    def test_context_manager_stops_process(self):
+        """
+        Verifies context manager stops process when exiting "with" block
+        """
+        process = SimpleProcess(shape=(2, 2))
+        self.stoppable = process
```
mathisrichter, line 69: Please consider declaring the `self.stoppable` in the `setUp()` method. Maybe something like this:
```python
import typing

class Stoppable(typing.Protocol):
    def stop() -> None:
        ...

def setUp(self) -> None:
    self.stoppable: Stoppable = None
```

tests/lava/magma/runtime/test_context_manager.py
```diff
@@ -0,0 +1,114 @@
...
+            sleep(2)
+
+        self.assertFalse(process.runtime._is_running)
+        self.assertFalse(process.runtime._is_started)
+
+    def test_context_manager_stops_runtime(self):
+        """
+        Verifies context manager stops runtime when exiting "with" block
+        """
+        self.process = SimpleProcess(shape=(2, 2))
```
mathisrichter, line 86: In the other test, `process` was a local variable, not a member `self.process`. Why use a member here?

mathisrichter (changes requested):
Hi @Gavinator98 

This looks very nice. Thanks for your work on this feature!

You are testing this feature with the `RunContinuous` run condition only, which uses a non-blocking call by default. Could you please also add tests for the `RunSteps` condition? One with `RunSteps(num_steps=..., blocking=True)` (the default) and one with `RunSteps(num_steps=..., blocking=False)`. It should be fine to only test these variants on the Process. You do not have to also test them for the runtime.

I also added a few minor suggestions. Could you please address those as well?

Thanks!

src/lava/magma/runtime/runtime.py
mathisrichter: Please check throughout, I did not mark this everywhere.

Gavinator98:
> Hi @Gavinator98
> 
> This looks very nice. Thanks for your work on this feature!
> 
> You are testing this feature with the `RunContinuous` run condition only, which uses a non-blocking call by default. Could you please also add tests for the `RunSteps` condition? One with `RunSteps(num_steps=..., blocking=True)` (the default) and one with `RunSteps(num_steps=..., blocking=False)`. It should be fine to only test these variants on the Process. You do not have to also test them for the runtime.
> 
> I also added a few minor suggestions. Could you please address those as well?
> 
> Thanks, Mathis

Thanks for the feedback, I'll try and address your comments early next week. One question, most of the test code is derived from [/tests/lava/magma/runtime/test_run_continuously_and_pause.py](https://github.com/lava-nc/lava/blob/main/tests/lava/magma/runtime/test_run_continuously_and_pause.py) Would it make sense to refactor SimpleProcess, etc for both into a utils module?

mathisrichter:
Sure, removing code duplication of unit tests is a valuable goal. Go for it and refactor shared code into a module but make sure you do not introduce dependencies between unit tests, where the behavior of one test is influenced by another.
It may also be worth renaming the Processes and ProcessModels. I feel that `SimpleProcess` is not very descriptive and could just be named `Process` or something short that explains what it is for.

You will notice that we have a lot of code duplication in unit test infrastructure throughout Lava tests... Maybe you will set a good example here that others will follow. Thanks for your initiative!

tests/lava/magma/runtime/test_context_manager.py
Gavinator98: This originated from an older version of the test where `self.stoppable` was `self.process`. As it is no longer needed I updated `process` to be a local variable.

Gavinator98:
> 

On further review, a majority of the tests have similar classes to `SimpleProcess` with slightly different names, variables, and number of processes. I don't think it makes sense to make that change anymore since it would only be shared by two of the test classes.

mathisrichter (approved):

mathisrichter:
Thanks, @Gavinator98 for this great new addition to Lava! If you are interested in contributing further in some form, do not hesitate to reach out. :)

Issue #606, opened 2023-02-06
Graded spike payload in pre-synaptic trace updates - `floating-pt` and `bit-approximate-loihi`


gkarray:
### User story
As a user of the learning API, I would like to use graded spike payloads to update pre-synaptic traces in order to implement more complex learning algorithms.  Especially, this is critical for prototype-based learning approaches.

Note: This story is for `floating-pt` and `bit-approximate-loihi` LearningDense PyProcessModels only.

### Conditions of satisfaction
- Works in simulation (`floating-pt` and `bit-approximate-loihi`)
- When there is a graded pre-synaptic spike on a learning-enabled synapse, this graded value could overwrite or be added to the pre-synaptic trace.
- The test coverage of the newly added module is >90%.
- The code has been reviewed and is merged into `main`.

### Acceptance tests
- Test that the pre-trace value is updated to or by the graded spike value after the pre-synaptic graded spike
- Test that pre-synaptic trace still decays exponentially with the given time constant
- Test that this pre-synaptic trace can be used in the learning engine as usual

### Tasks
- [x] Implement the feature on CPU backend (floating-point, bit-approximate)
- [x] Write unit tests
- [x] Update documentation (GK, 2h)
- [ ] Get 2 reviews (PW, EH, PP/MR?), merge into main (4h)

PR #607, opened 2023-02-06
Graded spike payload in pre-synaptic trace updates - `floating-pt` and `bit-approximate-loihi`

gkarray:
Issue Number: #606

Objective of pull request: As a user of the learning API, I would like to use graded spike payloads to update pre-synaptic traces in order to implement more complex learning algorithms. Especially, this is critical for prototype-based learning approaches.

## Pull request checklist

## Pull request type

 

 

Please check your PR type:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] Documentation changes

- Currently, feature described by the user story is not present.

## What is the new behavior?

- The feature described by the user story is present.

- [ ] Yes
- [x] No


src/lava/magma/core/learning/constants.py
```diff
@@ -85,3 +86,10 @@
     str_symbols.Y2: 1,
     str_symbols.Y3: 2
 }
+
+
+class GradedSpikeCfg(IntEnum):
+    DEFAULT = 0
```
mathisrichter, line 92: I wonder if this should have a descriptive name also, for example, `UPDATE`. The fact that it is the default should be reflected in the default value where it is instantiated.

src/lava/magma/core/model/py/connection.py
```diff
@@ -631,36 +635,80 @@ def _init_randoms(self) -> None:
...
+        trace x1 by payload/2, upon spiking.
+        (2) GradedSpikeCfg.ADD_SATURATION adds payload/2 to the pre-synaptic
+        trace x1, upon spiking, saturates x1 to 127.
+        (3) GradedSpikeCfg.ADD_NO_SATURATION adds payload/2 to the pre-synaptic
+        trace x1, upon spiking, keeps only overflow from 127 in x1,
+        adds regular impulse to x2 on overflow.
+
+        Within-epoch spike times are recorded.
+        With GradedSpikeCfg.ADD_NO_SATURATION, only spike times of spikes
+        triggering x1 overflow and x2 impulse addition are recorded.
```
mathisrichter, line 654: This docstring is copied to several places within this file. Can that be defined once in a central location and just referenced elsewhere?

src/lava/magma/core/model/py/connection.py
```diff
@@ -726,7 +774,7 @@ def _compute_trace_histories(self) -> typing.Tuple[np.ndarray, np.ndarray]:
 
         # most naive algorithm to decay traces
         # TODO decay only for important time-steps
```
mathisrichter, line 776: See comment below

src/lava/magma/core/model/py/connection.py
```diff
@@ -726,7 +774,7 @@ def _compute_trace_histories(self) -> typing.Tuple[np.ndarray, np.ndarray]:
 
         # most naive algorithm to decay traces
```
mathisrichter, line 775: Is this a hidden todo? If so, create an issue for developing a more sophisticated algorithm. If not, just remove the comment; also in the other place where it showed up, below.

tests/lava/proc/dense/test_learning.py
```diff
@@ -0,0 +1,1649 @@
...
+
+        return pattern_pre, learning_dense, pattern_post
+
+    expected_x1_data_default = \
+        [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 12.0, 12.0, 12.0, 12.0, 24.0, 24.0,
+         24.0, 24.0, 19.0, 19.0, 19.0, 19.0, 27.0, 27.0, 27.0, 27.0, 34.0, 34.0,
+         34.0, 34.0, 25.0, 25.0, 25.0, 25.0, 19.0, 19.0, 19.0, 19.0, 15.0, 15.0,
+         15.0, 15.0, 26.0, 26.0, 26.0, 26.0, 20.0, 20.0, 20.0, 20.0, 16.0, 16.0,
+         16.0, 16.0, 12.0, 12.0, 12.0, 12.0, 24.0, 24.0, 24.0, 24.0, 18.0, 18.0,
+         18.0, 18.0]
```
mathisrichter, line 953: Where do these values come from? How do you know that these are expected? Can we make that more explicit so that the test code becomes more readable to people?

Do we need 63 time steps to make sure the algorithm is working as expected? If it were just 5, for instance, the difference between the different modes would be more obvious - I guess.

mathisrichter (commented):
I have not reviewed this to a point, where I could approve or block it, so I am just leaving comments.

It all looks good though, at least from a high level coding-style point of view. I only added a few minor remarks.

The one thing we could maybe improve are the known-value tests. It is not clear where the known values come from or how we know they are correct. There are also so many that the significance of individual numbers is unclear. Please keep in mind that someone will at some point look at this code and need to understand why all these values as deemed correct. Make sure that it is very easy to understand that from the code. That person could be yourself one year from now. :)

tests/lava/proc/dense/test_learning.py
```diff
@@ -0,0 +1,1649 @@
...
+        [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 12.4608125, 12.4608125, 12.4608125,
+         12.4608125, 24.9285623, 24.9285623, 24.9285623, 24.9285623, 17.8620954,
+         17.8620954, 17.8620954, 17.8620954, 26.3424583, 26.3424583, 26.3424583,
+         26.3424583, 32.4189037, 32.4189037, 32.4189037, 32.4189037, 23.2291596,
+         23.2291596, 23.2291596, 23.2291596, 16.6444202, 16.6444202, 16.6444202,
+         16.6444202, 11.9262482, 11.9262482, 11.9262482, 11.9262482, 23.2662409,
+         23.2662409, 23.2662409, 23.2662409, 16.6709901, 16.6709901, 16.6709901,
+         16.6709901, 11.9452864, 11.9452864, 11.9452864, 11.9452864, 8.5591717,
+         8.5591717, 8.5591717, 8.5591717, 22.1329145, 22.1329145, 22.1329145,
+         22.1329145, 15.8589262, 15.8589262, 15.8589262, 15.8589262]
```
mathisrichter, line 103: Same here, where do these values come from?

tests/lava/proc/dense/test_learning.py
gkarray: These values come from the simulation itself. They have been thoroughly validated (in the sense that they approximate the chip behavior). The purpose of these known value tests is to help us notice breaking changes in the future (which would make these tests fail.

As for the second question, the choice of 63 is arbitrary here. It could have been something else. It is true that the fact that number is high makes the difference between the different modes less obvious to the user/developer reading this code snippet.

However, the results we are testing for here depend not only on the chosen mode, but also on:
1. The learning rule condition (for each mode, all three different learning rule conditions are tested)
2. The epoch length (set to 4 in all these tests)
3. Pre-synaptic spiking history
4. Pre-synaptic within-epoch spike times
5. Post-synaptic spiking history (in some cases)

The purpose of having a high enough number of time steps is for making the pre/post-synaptic spiking history and pre-synaptic within epoch spike times more "diverse", so that we can cover more cases in each test.

src/lava/magma/core/model/py/connection.py
gkarray: Addressed this.

src/lava/magma/core/learning/constants.py
gkarray: Addressed this.

src/lava/magma/core/learning/constants.py
```diff
@@ -85,3 +86,10 @@
...
+    OVERWRITE = 1
+    ADD_SATURATION = 2
+    ADD_NO_SATURATION = 3
```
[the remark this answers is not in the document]
gkarray: Addressed this.

src/lava/magma/core/model/py/connection.py
```diff
@@ -1191,7 +1290,7 @@ def _compute_trace_histories(self) -> typing.Tuple[np.ndarray, np.ndarray]:
...
         # TODO decay only for important time-steps
```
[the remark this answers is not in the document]
gkarray: Addressed this.

src/lava/magma/core/model/py/connection.py
```diff
@@ -1114,37 +1181,69 @@ def _store_impulses_and_taus(self) -> None:
...
+            self.x1[spiked] = scaled_activations[spiked]
+
+        elif self._graded_spike_cfg == GradedSpikeCfg.ADD_SATURATION or \
+                self._graded_spike_cfg == GradedSpikeCfg.ADD_NO_SATURATION:
+            sums = self.x1 + scaled_activations
+
+            # if self._graded_spike_cfg == GradedSpikeCfg.ADD_NO_SATURATION:
+            #     overflow_idx = sums > 127
+            #     update_t_spike = update_t_spike & overflow_idx
+            #     x2_update_idx = x2_update_idx & overflow_idx
```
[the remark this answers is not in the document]
gkarray: Addressed this.

src/lava/magma/core/model/py/connection.py
gkarray: It is actually similar but not exactly the same within this file.
There is a subtle difference between the FloatingPoint and the BitApprox versions when it comes to this docstring.

An option would be to define this docstring in the PyLearningConnection class (which is the parent class of both the FloatingPoint and BitApprox versions) might be an idea. However, the method in question (`_process_pre_spikes`) is abstract in PyLearningConnection and is only implemented in children classes, which makes me a bit reluctant to put the docstring there.

PhilippPlank (approved):
This concept of how the payload modifies the traces seems quite complicated. We should have a good example in the learning tutorial to show what will happen.
If there is no issue for that yet, please create one.

tests/lava/proc/dense/test_learning.py
mathisrichter: I was hoping we could design simple tests that cover all the features and are still understandable to people. Then you could actually write out the expected numbers by hand, explain why you expect them in the test, and anyone reading the tests later on could have a chance at understanding why those numbers are expected.

Currently, if the test suddenly fails, the only thing a developer can understand is that the values are different. He or she cannot see why they should be what you expect unless they go through all the learning engine code in detail and understand all of that.

If the tests depend on these five factors you are listing, maybe there should really be a test for each of these factors in isolation.

gkarray: Makes perfect sense. Addressed this in a new commit.

mathisrichter (approved):

Issue #362, opened 2022-09-22
Raster Plot for spiking activity


ackurth-nc:
**Lava version:**

- [ ] **0.5.0** (feature release)
- [ ] **0.4.1** (bug fixes)
- [x] **0.4.0** (current version)
- [ ] **0.3.0**
- [ ] **0.1.2**

**I'm submitting a ...** 

- [ ] bug report
- [x] feature request
- [ ] documentation request

**Feature request:**
Function to plot spiking activity of raster plot akin to one implemented in #309

PR #623, opened 2023-02-14
Integrate raster plot

ssgier:
Issue Number: #362

Objective of pull request: Integrate raster plot

## Pull request checklist

## Pull request type

 

 

Please check your PR type:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] Documentation changes

- Code to generate raster plot inlined in tutorial notebook

## What is the new behavior?

- Code to generate raster plot exposed in a new module: lava.utils.plots

- [ ] Yes
- [x] No

## Supplemental information

- A few tests fail locally, but they do as well on main. I made sure the same subset of tests are failing and that they are unrelated to my change. The last cause is sqlite3.OperationalError: unable to open database file. I am running pytest with Python 3.10.9 under Linux. Is this issue known? If not, perhaps best to discuss separately. I will also take a look into what might be the root cause.
- I did not re-evaluate the notebook to make the review easier. However, I suggest that I rerun it eventually, for consistency.
- The assert pasted in the issue asserts that the stride is less than the number of time steps. But the way I read it, it should be less than the number of neurons. I changed it accordingly. Let me know if I misunderstood this. Also, I replaced the assert with raising a ValueError to placate the linter.
- I am currently not including a unit test because the new method just creates a plot. Let me know if it should still be tested in some way.


src/lava/utils/plots.py
```diff
@@ -0,0 +1,45 @@
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+# See: https://spdx.org/licenses/
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+
+def raster_plot(spks, stride=6, fig=None, color='b', alpha=1):
```
mathisrichter, line 9: Please add type annotations here using the `typing` module, for instance
```python
from typing import Optional

def raster_plot(spks: np.ndarray, stride: Optional[int] = 6, ...) -> <ReturnType>:
```

src/lava/utils/plots.py
```diff
@@ -0,0 +1,45 @@
...
+    """Generate raster plot of spiking activity.
+
+    Parameters
+    ----------
+    spks : np.ndarray shape (num_neurons, num_timesteps)
+        Spiking activity of neurons, a spike is indicated by a one
+    stride : int
+        Stride for plotting neurons
+    """
```
mathisrichter, line 18: Please document all available parameters of the function.

src/lava/utils/plots.py
```diff
@@ -0,0 +1,45 @@
...
+import numpy as np
+import matplotlib.pyplot as plt
+
+
+def raster_plot(spks, stride=6, fig=None, color='b', alpha=1):
+    """Generate raster plot of spiking activity.
+
+    Parameters
+    ----------
+    spks : np.ndarray shape (num_neurons, num_timesteps)
```
mathisrichter, line 14: Could we rename `spks` to `spikes`?

mathisrichter, line 14: Could you please remove everything after `np.ndarray` here. Instead, please incorporate a full but concise description of the expected format of the `spikes` parameter in the description (next line).

src/lava/utils/plots.py
```diff
@@ -0,0 +1,45 @@
...
+        Spiking activity of neurons, a spike is indicated by a one
+    stride : int
+        Stride for plotting neurons
+    """
+    num_neurons = spks.shape[0]
+    num_time_steps = spks.shape[1]
```
mathisrichter, line 20: Please add input validation here, where users receive meaningful error messages when `spikes` is specified in the wrong format.

Similarly, please add input validation for the remaining parameters.

I suggest writing unit tests for these cases that check whether the exceptions are raised in exactly the right circumstances.

src/lava/utils/plots.py
```diff
@@ -0,0 +1,45 @@
...
+
+    if stride >= num_neurons:
+        raise ValueError("Stride must be less than the number of neurons")
```
mathisrichter, line 23: This should be unit tested.

src/lava/utils/plots.py
```diff
@@ -0,0 +1,45 @@
...
+    plt.xlabel('Time steps')
+    plt.ylabel('Neurons')
+
+    for i in range(0, num_neurons, stride):
+        spike_times = time_steps[spks[i] == 1]
+        plt.plot(spike_times,
+                 i * np.ones(spike_times.shape),
+                 linestyle=' ',
+                 marker='o',
+                 markersize=1.5,
```
mathisrichter, line 41: It may be nice to expose these parameters to the user as optional parameters to the function.

src/lava/utils/plots.py
```diff
@@ -0,0 +1,45 @@
...
+
+    time_steps = np.arange(0, num_time_steps, 1)
+    if fig is None:
+        fig = plt.figure(figsize=(10, 5))
+
+    plt.xlim(-1, num_time_steps)
+    plt.yticks([])
+
+    plt.xlabel('Time steps')
+    plt.ylabel('Neurons')
```
mathisrichter, line 33: It may be nice to expose these labels to the user as optional parameters to the function.

src/lava/utils/plots.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    num_neurons = spks.shape[0]
+    num_time_steps = spks.shape[1]
+
+    if stride >= num_neurons:
+        raise ValueError("Stride must be less than the number of neurons")
+
+    time_steps = np.arange(0, num_time_steps, 1)
+    if fig is None:
+        fig = plt.figure(figsize=(10, 5))
```
mathisrichter, line 27: It may be nice to expose the figure size parameters to the user as optional parameters of the function.

mathisrichter (changes requested):
Hi @ssgier 
Thanks for your first contribution to Lava! I'm glad to see that you dug right in and created a PR. This is a good first start; the code that we had pasted in the issue was just meant as a starting point. I believe we could make this more user friendly by exposing more parameters to the user, making the function more flexible. I would also suggest to add input validation to catch errors early and provide meaningful feedback to users when they make a mistake. This should then be unit tested to ensure that the validation works as expected.

mathisrichter:
Regarding your notes:

> A few tests fail locally, but they do as well on main. I made sure the same subset of tests are failing and that they are unrelated to my change. The last cause is sqlite3.OperationalError: unable to open database file. I am running pytest with Python 3.10.9 under Linux. Is this issue known? If not, perhaps best to discuss separately. I will also take a look into what might be the root cause.

I am not aware of this issue. Could you look through the issues in Github to see if this has been filed? If not, please create a separate issue. This could be an environment problem.

> I did not re-evaluate the notebook to make the review easier. However, I suggest that I rerun it eventually, for consistency.

Agreed. Re-evaluate it once you think the notebook iteself no longer needs to be changed.

> The assert pasted in the issue asserts that the stride is less than the number of time steps. But the way I read it, it should be less than the number of neurons. I changed it accordingly. Let me know if I misunderstood this. Also, I replaced the assert with raising a ValueError to placate the linter.

Yes, I believe you are correct. Good catch. :)

> I am currently not including a unit test because the new method just creates a plot. Let me know if it should still be tested in some way.

Best practice is to write tests for every qualitatively different behavior of a piece of code.
Things to test here are, for instance:
- that a figure object is returned with default parameters
- that a meaningful error is raised for all the invalid inputs you are handling (I suggested adding more)
- that no errors are raised when parameters are valid

src/lava/utils/plots.py
ssgier: I added some validation. But there are two kinds that I left out for now:
- Validation of the parameters directly related to matplotlib, since we are merely forwarding them. In this case we might want to rely on the matplotlib-internal validation. Let me know what you think.
- Type checking validation: I saw that there are some cases in this project where this is done and type errors are raised. Is there a convention?

PhilippPlank (approved):

src/lava/utils/plots.py
mathisrichter: I agree with leaving out both types of validation. Type checking is sometimes considered valuable by some but I tend to guide users with type annotations rather than add a lot of validation. That way, they can at least get some warning highlights in the IDE.

tests/lava/utils/test_plots.py
```diff
@@ -0,0 +1,68 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+# See: https://spdx.org/licenses/
+
+import unittest
+import numpy as np
+import matplotlib.pyplot as plt
+from lava.utils.plots import raster_plot
+
+np.random.seed(0)
+_SPIKES = np.random.randint(2, size=(10, 20))
```
mathisrichter, line 11: You could also set this up like this
```python
class TestInputValidation(unittest.TestCase):
    def setUp(self) -> None:
        """This will be executed to set up every test."""
        self.spikes = np.random.randint(2, size=(10, 20))

    def test_return_figure_on_valid_input(self) -> None:
        fig = raster_plot(self.spikes)
        ...
```

mathisrichter (changes requested):
Looks very nice, @ssgier !
I remember you still wanted to reevaluate the Jupyter notebook? We can merge the PR once that is done.

tests/lava/utils/test_plots.py
ssgier: Thanks for reviewing @mathisrichter! On the above: the main difference I can see would be that ```spikes``` would get a different value for each sub-test. Is there value in this or is this proposed change for improving the code style?

mathisrichter: This was rather a point on code style. This may be personal taste - I prefer such member variables over variables in the module scope.

If you want the same values for each test, you could set the seed in the `setUp` method as well.

mathisrichter (approved):

ssgier:
Following up on the locally failed unit tests: I had a look and there were actually two separate root causes:
1. On my system, the ulimit suggested in the documentation (4096) was not high enough. Pytest opened over 11K files during the run, a lot of them shared memory files.
2. The mnist tutorial test attempted to load a weights file which is stored with Git LFS.

Setting a much higher ulimit and installing Git LFS resolved the issues and made all tests pass. I am not running an officially supported Linux distribution, so I'm not sure if the problems can also arise with the supported OSes.

mathisrichter:
Thanks for following up on that.

The first has been a problem for a while - this is due to a problem with the `multiprocessing` library we are using, I believe. We are working on exchanging that, which will hopefully resolve that issue.

The second could just be missing documentation about having to use Git LFS to get the files. Feel free to add that to the documentation where you think it would be appropriate.

mathisrichter:
You PR is now merged. Thanks again for your work on this! We very much hope you will continue to support Lava with new features and bug fixes soon! :)

PR #624, opened 2023-02-15
Synaptic delays for Dense connections

PhilippPlank:
Issue Number: #237 
Based on PR: #267 

Objective of pull request: As a user, I want to be able to specify synaptic delays in the Dense Process to enable algorithmic work.

## Pull request checklist

## Pull request type

 

 

Please check your PR type:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] Documentation changes

-

## What is the new behavior?

-

- [ ] Yes
- [x] No


src/lava/proc/dense/models.py
```diff
@@ -188,3 +188,148 @@ def run_spk(self):
...
 
         self.recv_traces(s_in)
+
+
+class AbstractPyDelayDenseModel(PyLoihiProcessModel):
+    """Abstract Conn Process with Dense synaptic connections which incorporates
+    delays into the Conn Process.
+    """
+
+    def get_del_wgts(self):
```
bamsumit, line 198: Can we add a test for this weight & delay matrix expansion to tensor?

src/lava/proc/dense/process.py
```diff
@@ -144,3 +144,83 @@ def __init__(self,
                          log_config=log_config,
                          learning_rule=learning_rule,
                          **kwargs)
+
+
+class DelayDense(Dense):
```
bamsumit, line 149: I am thinking about the proliferation of all different variants. It might make the end user's life simpler if we check for non-zero delay and switch to this delayed computation. I will leave this decision up to you.

src/lava/proc/dense/process.py
```diff
@@ -144,3 +144,83 @@ def __init__(self,
...
+                         log_config=log_config,
+                         **kwargs)
+
+        self._validate_delays(weights, delays)
+        max_delay = int(np.max(delays))
+        shape = weights.shape
+
+        # Variables
+        self.delays = Var(shape=shape, init=delays)
+        self.a_buff = Var(shape=(shape[0], max_delay + 1) , init=0)
```
bamsumit, line 215: It might be useful to add `max_delay` as a parameter rather than using the max delay value that we get during initialization. This works for inference, but if the user pauses the process and sets a new delay matrix with a larger delay, it might cause errors.

src/lava/proc/dense/process.py
```diff
@@ -144,3 +144,83 @@ def __init__(self,
...
+
+    @staticmethod
+    def _validate_delays(weights: np.ndarray, delays: np.ndarray) -> None:
```
bamsumit, line 218: Can we also make sure the delay values are integers? I am fine with raising an error in process or flooring it down in the process model.

bamsumit (changes requested):
Thanks @PhilippPlank for taking care of this. It looks nice and clean. I have added some comments for your consideration.

tests/lava/proc/dense/test_models.py
```diff
@@ -553,3 +553,484 @@ def test_bitacc_pm_recurrence(self):
...
+        rcfg = DenseRunConfig(select_tag='floating_pt')
+        dense.run(condition=rcnd, run_cfg=rcfg)
+        # Gather spike data and stop
+        spk_data_through_run = spr.spk_data.get()
+        dense.stop()
+        # Gold standard for the test
+        # a_out will be equal to 1 at timestep 7, because the dendritic
+        #  accumulators work on inputs from the previous timestep + 2.
+        expected_spk_data = np.zeros((num_steps, shape[0]))
+        expected_spk_data[6, 2] = 1.
```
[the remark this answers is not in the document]
PhilippPlank: The equation you stated is not correct, since the sum is 7 :P
We send at time step 4, at time step 5 the spikes arrive without delay. 
Not sure what the dend buffer delay would be, I just added the delay to the expected values from the dense unit tests.

src/lava/proc/dense/process.py
PhilippPlank: We decided to go this route for now, since we did the same with learning. Delays and learning are also not compatible. Figuring out how we compose processes and provide a simple, but powerful user interface is an ongoing discussion.

src/lava/proc/dense/models.py
```diff
@@ -188,3 +188,148 @@ def run_spk(self):
...
+    def run_spk(self):
+        # The a_out sent on a each timestep is a buffered value from dendritic
+        # accumulation at timestep t-1. This prevents deadlocking in
+        # networks with recurrent connectivity structures.
+        self.a_out.send(self.a_buff[:, 0])
+        if self.num_message_bits.item() > 0:
+            s_in = self.s_in.recv()
+        else:
+            s_in = self.s_in.recv().astype(bool)
+        self.update_act(s_in)
```
gkarray, line 271: Looks like `update_act` is called every time step, which in turn calls `calc_act`, which in turn calls `get_del_wgts`.
Couldn't we execute `get_del_wgts` just one time at initialization and store its result for later use ?
The computation happening there seem to be relatively heavy and might hinder performance if executed every time step.

gkarray (approved):
I left a single comment about potential performance improvement. Might even be not that critical if the overhead incurred there is not important.
Otherwise looks good! I'll approve to not block PR.

src/lava/proc/dense/models.py
PhilippPlank: I actually tried that, but there are some arguments why this is not trivial to change. First, at initialization, the wgts and dlys are not numpy arrays yet, but LavaPyTypes - no operations work yet - this will change when the proc model is actually executed. More important though is the fact, that the wgts and maybe even the delays could change during execution. 

For now this seems to be the robust way to process delays.

src/lava/proc/dense/process.py
PhilippPlank: Good suggestion, I added `max_delay` as a parameter.

src/lava/proc/dense/process.py
PhilippPlank: Included a check that the dtype of a matrix is int. The single value already needed to be an int. Also added unit tests.

src/lava/proc/dense/models.py
PhilippPlank: Done. Also made the method a static method.

tests/lava/proc/dense/test_models.py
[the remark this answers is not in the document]
bamsumit: Ah that's correct.

Issue #642, opened 2023-02-23
File descriptors leaking with multiprocessing lib (error: "Too many open files")


ssgier:
**Note**
This problem is already known, but I could not find an existing issue on GitHub. I am creating this issue so I can submit a PR for a fix.

**Describe the bug**
Resource handles aren't closed properly in several places where the Python multiprocessing library is used:
- shared memory files
- pipes
- child processes

**To reproduce current behavior**
When I run pytest on a Linux with ulimit 1024 (default on my machine), file descriptors are leaking, eventually leading to a "Too many open files" error. This behavior is known and documented.

**Expected behavior**
File descriptors shouldn't leak. Setting a higher ulimit shouldn't be necessary to make the tests pass.

**Environment (please complete the following information):**
 - Device: desktop computer
 - OS: Linux
 - Lava version [e.g. 0.6.0]

**Additional context**
The leak can be fixed by properly closing the resources in each of the above mentioned cases.

PR #643, opened 2023-02-23
Plug file descriptor leaks

ssgier:
Issue Number: #642 

Objective of pull request: fix file descriptor leaks.

## Pull request checklist

## Pull request type

 

 

Please check your PR type:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] Documentation changes

- File descriptors are leaking in multiple places. Setting high ulimit is necessary for successful pytest run.

## What is the new behavior?

- File descriptors are not leaking anymore. Setting ulimit is not necessary for successful pytest run anymore.

- [ ] Yes
- [x] No

## Supplemental information

Key points all involve addressing sub-optimal usage of the Python multiprocessing library:
- ensuring close is called on shared memory file handles (not the same as unlink, which is invoked by the shutdown of the manager)
- closing pipes
- closing resource handles to child processes

**Note**: the fix works for Linux. For other OSes it would have to be checked as well.


mgkwill:
Thanks for this PR/bug fix. I'll take a look tomorrow and give more feedback, in the mean time I've added a few reviewers.

mgkwill:
Thanks for this PR/bug fix. I'll take a look tomorrow and give more feedback, in the mean time I've added a few reviewers.

src/lava/magma/runtime/message_infrastructure/close_on_shutdown_smm.py
```diff
@@ -0,0 +1,26 @@
...
+    def start(self):
+        self._smm.start()
+
+    def shutdown(self):
+        for shm in self._shms:
+            shm.close()
+        self._shms.clear()
+        self._smm.shutdown()
+
+    def shared_memory(self, size: int):
```
mathisrichter, line 23: Missing return type annotation

src/lava/magma/runtime/message_infrastructure/close_on_shutdown_smm.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: LGPL 2.1 or later
+# See: https://spdx.org/licenses/
+
+
```
mathisrichter, line 5: Please remove one of these empty lines.

src/lava/magma/runtime/message_infrastructure/close_on_shutdown_smm.py
```diff
@@ -0,0 +1,26 @@
...
+from multiprocessing.managers import SharedMemoryManager
+
+
+class CloseOnShutdownSMM:
```
mathisrichter, line 9: What do you think about renaming this to `SharedMemoryManager`? The fact that it closes on shutdown is not too important for the user to worry about, I think. It is rather strange that multiprocessing does not do that for you.
For the code below, you could then use
```python
import multiprocessing as mp
self._smm = mp.managers.SharedMemoryManager()
```

src/lava/magma/runtime/message_infrastructure/close_on_shutdown_smm.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self):
+        self._shms = []
```
mathisrichter, line 11: I tend to prefer readable/pronouncable variable names like
```python
self._managers = []
self._manager = mp.managers.SharedMemoryManager()
```

ssgier:
Thanks for reviewing @mathisrichter!

I will implement your suggestions during the coming days and will also think of a way to write some specific tests.

### Some follow-up information
Below is a screenshot of my terminal emulator, showing proof of concept:
Screenshot_2023-02-24_19-52-56
This does the following:
* define an alias to run test_learning_rule and filter for relevant output
* set a low ulimit for open file descriptors
* run test with main branch -> fails with "Too many open files"
* run test with fix branch -> passes
* set higher ulimit for open file descriptors
* run test with main branch -> passes now

This means that with low ulimit, only the fix branch avoids the files error, but with high ulimit, both avoid it. This result is deterministic. It can be run any number of times with the same result.

I am running:
- Python 3.10.9
- Linux kernel 6.1.12

**However**: I tested the fix today on my MacBook (macOS Ventura 13.2) and there it does not work. After going back to the Linux machine and taking a deeper look, I noticed that if I call the tests in a specific way (by repeating the same test a large number of times), some descriptors still leak, although much more slowly, eventually leading to the same error.

**Conclusion**: This change fixes some leaks, but not all of them yet. At least one remains. I will try to track them all down and come back with an update soon.

mathisrichter:
@ssgier Thanks for that deeper analysis! I brought up your PR internally and learned that there may be tests for this error in some branch. I asked @joyeshmishra to add them as a comment to this thread.
But I don't know the details of what those tests cover, so in the meantime, please continue with your own investigation.

ssgier:
**Update**: Found some more leaks and committed a draft of a fix here.

**New behavior**: With these fixes applied, all file descriptor leaks appear to be resolved on both Linux and macOs and the whole test suite runs through with low ulimit.

**Next steps**: I will implement @mathisrichter's suggestions and also write an explicit test. It turns out that there is a simple way of doing this:
```
import psutil
p = psutil.Process()
num_fds_before = p.num_fds()

<run subject under test>

num_fds_after = p.num_fds()
assert num_fds_after == num_fds_before
```

**Details**
1. The shared memory manager instance opens pipes internally, which are only closed when the Python object is reclaimed.
2. Some of the ports are never joined by the runtime.
3. This one is specific to macOS: the multiprocessing semaphore opens a special file descriptor: the POSIX Semaphore (type PSXSEM when running ```lsof```). This file descriptor is only closed when the Python object is reclaimed.
4. Threads were leaking in pypychannel. They waited on the semaphore and never got to check the self._done condition. This is not a file descriptor leak, but it is related to fixing item 3, because the threads have a reference to the semaphore object. Cleanest would be to also join the threads, but this slows down the stopping process of the runtime.

I tried to keep the fixes as non-intrusive as possible. One shortcoming is that some of them rely on objects being reclaimed, but it works in practice and from what I can see, it does not break existing functionality.

ssgier:
Looks like the outcome of test_py_in_port_probe has become unpredictable. I will try to reproduce locally and fix.

mathisrichter (approved):

src/lava/magma/runtime/runtime.py
```diff
@@ -407,11 +415,11 @@ def stop(self):
 
     def join(self):
         """Join all ports and processes"""
-        for port in self.runtime_to_service:
```
joyeshmishra, line 410: I see you have created a new list open_ports but have not eliminated the old lists (runtime_to_service and service_to_runtime). If there are no new items, and the two old containers are by convention, even if you delete the references from open ports, the references are still there in the old lists. So I don't get what the open ports is accomplishing here?

joyeshmishra (changes requested):
Thanks for this contribution. Looks great. Expecting you have verified on all three platforms before merge. Small comment on code otherwise looks great,

src/lava/magma/runtime/runtime.py
ssgier: * open_ports is only to track ports that need to be joined when the runtime is stopped. The join operation in the ports implements clean up logic needed to prevent leaking. This is not about deleting references to the ports, as they don't themselves implement resource clean up in their destructors (they aren't RAII objects).
* runtime_to_service and service_to_runtime looked like they were also intended to track ports to be joined, but they also serve other purposes.
* by introducing open_ports, I merely created a separate variable for the tracking of ports, while leaving the remaining accounting with runtime_to_service and service_to_runtime.
* The reason for splitting into a separate variable was that the the service_to_runtime and runtime_to_service were missing part of the ports to track.

ysingh7 (approved):

ssgier:
@joyeshmishra thanks for reviewing! I did verification on Linux and macOS but not on Windows. Is resource leakage also a problem on Windows? It looked like a problem specific to Unix-like systems.

ssgier:
As discussed separately with @mathisrichter, removed the now obsolete hint from README and tutorial. The tag referred in the README would have to be updated in the next release.

harryliu-intel (approved):
Looks good to me, thanks.

src/lava/magma/runtime/runtime.py
mathisrichter: @ssgier Could you please address this question before we merge the PR?

ssgier: My message above was intended to address this. Is anything unclear?

mathisrichter: No sorry, I had apparently not reloaded the browser when I wrote my message and I had not seen your reply. My bad. Your answer sounds fine to me.

mathisrichter: @joyeshmishra Could you please review this and merge the PR if everything is in order?
Things look good to me.

joyeshmishra: Looks good to me. Please go ahead.

joyeshmishra (approved):

Issue #282, opened 2022-07-25
Refractory period for LIF Process


mathisrichter:
### User story

As a user, I want the LIF Process to support a refractory period after spiking.

Note: A side effect of this missing feature is that the current Python implementation resets the voltage to 0 after a spike, whereas on the hardware it is reset to 128 (which implements the refractory period). The original story was a bug around that.

### Conditions of satisfaction
- to be determined

### Acceptance tests
- to be determined

GitUser369:
I believe that this is just a reporting difference. The internal (hidden) states should be the same. You can confirm this by looking at when subsequent spikes occur.

mathisrichter:
The state of the voltage on Loihi is in fact 128 after a spike. As you are correctly pointing out, that initial voltage does not have an impact on its later evolution.

I think this behavior should still be reflected in the Python ProcessModel to enable easy comparison between results from hardware and simulation.

mathisrichter:
This story is really about adding support for a refractory period in the LIF neuron. The different behavior in the bit-accurate Python ProcessModel is just an artifact of that missing feature. I have changed the story to reflect this.

tim-shea:
#655 was solved with a merged refractory process.

PR #655, opened 2023-03-21
LIF refractory floating point

ssgier:
Issue Number:  #282

Objective of pull request: Add LIF process with refractory period

## Pull request checklist

## Pull request type

 

 

Please check your PR type:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] Documentation changes

- No LIF process with refractory period is available

## What is the new behavior?

- LIF process with refractory period is available. CPU/Floating point process model that implements the process is available.

- [ ] Yes
- [x] No

## Supplemental information

As discussed with @mathisrichter, I created a new process based on the LIF process with reset. This change currently only includes the floating point process model in Python as a start. It could serve as a starting point to discuss/implement other process models.


mathisrichter:
Oh boy, sorry @ssgier for the long delay. I have not been monitoring Github for a while and we overlooked this PR.
I have now assigned a few people from my team to review it.

Hopefully, we can review and get this merged soon now.
My apologies!

src/lava/proc/lif/process.py
```diff
@@ -347,3 +347,75 @@ def __init__(
...
+            du=du,
+            dv=dv,
+            bias_mant=bias_mant,
+            bias_exp=bias_exp,
+            vth=vth,
+            name=name,
+            log_config=log_config,
+        )
+
+        if refractory_period < 1:
```
weidel-p, line 417: In my understanding, a refractory period of 0 should be valid and mean "no refractory period".

weidel-p (changes requested):
Thanks for this clean implementation there is not much to ask for.  

The only thing is, that I would like to change the default value to 0 instead of 1. In your current implementation, 1 means "no refractoryness" and yields the same results as the normal LIF. In my understanding, 1 should mean that the voltage is clamped to zero for one timestep after spiking. Hence I would change the line

```non_refractory = self.refractory_period_end <= self.time_step```

to 

``` non_refractory = self.refractory_period_end < self.time_step ``` 

and allow refractory_period to be 0 but not negative.

File: src/lava/proc/lif/process.py (hunk 1 above)
Comment (ssgier) [reply src/lava/proc/lif/process.py:417]: The reason why I wrote the check like this is that, as you pointed out, a refractory period of zero would result in the same behavior as the normal LIF, which might be unintended by the user (why would he not use the normal LIF instead?). Do you want me to adjust this nonetheless?

Comment (weidel-p) [reply src/lava/proc/lif/process.py:414]: no, I think that's fine. Thanks.

File: tests/lava/proc/lif/test_models.py [hunk 2]
```diff
@@ -824,3 +825,37 @@ def test_fixed_model(self):
...
+        v_logger = io.sink.Read(buffer=num_steps)
+        v_logger.connect_var(lif_refractory.v)
+
+        lif_refractory.run(condition=RunSteps(num_steps),
+                           run_cfg=Loihi2SimCfg(select_tag="floating_pt"))
+
+        v = v_logger.data.get()
+        lif_refractory.stop()
+
+        # voltage is expected to remain at reset level for two time steps
```
PhilippPlank, line 857: Please start comments with capital letters to stay consistent.

weidel-p (approved):

Issue #586, opened 2023-01-26
Execution time of `.read()` on a RefPort grows the more we call it


gkarray:
**Describe the bug**
Execution time of `.read()` on a RefPort grows the more we call it.

**To reproduce current behavior**
1. When I run this code ...
```python
# Imports
import numpy as np
import time

from lava.magma.core.process.process import AbstractProcess
from lava.magma.core.process.ports.ports import RefPort
from lava.magma.core.process.variable import Var

from lava.magma.core.sync.protocols.loihi_protocol import LoihiProtocol
from lava.magma.core.model.py.ports import PyRefPort
from lava.magma.core.model.py.type import LavaPyType
from lava.magma.core.resources import CPU
from lava.magma.core.decorator import implements, requires
from lava.magma.core.model.py.model import PyLoihiProcessModel

from lava.magma.core.run_conditions import RunSteps
from lava.magma.core.run_configs import Loihi1SimCfg


# MyProcess Process
class MyProcess(AbstractProcess):
    def __init__(self):
        super().__init__()
        self.var = Var(shape=(10, 10), init=np.zeros((10, 10)))


# MyProcess ProcessModel
@implements(proc=MyProcess, protocol=LoihiProtocol)
@requires(CPU)
class MyProcessPM(PyLoihiProcessModel):
    var: np.ndarray = LavaPyType(np.ndarray, float)


# Reader Process
class Reader(AbstractProcess):
    def __init__(self):
        super().__init__()
        self.ref_port = RefPort(shape=(10, 10))


# Reader ProcessModel
@implements(proc=Reader, protocol=LoihiProtocol)
@requires(CPU)
class ReaderPM(PyLoihiProcessModel):
    ref_port: PyRefPort = LavaPyType(PyRefPort.VEC_DENSE, float)

    def post_guard(self):
        return True

    def run_post_mgmt(self):
        checkpoint_1 = time.perf_counter()
        self.ref_port.read()
        checkpoint_2 = time.perf_counter()
        print(checkpoint_2-checkpoint_1)


def main():
    # Params
    num_steps = 10000
    run_cnd = RunSteps(num_steps=num_steps)
    select_tag = "floating_pt"
    run_cfg = Loihi1SimCfg(select_tag=select_tag)

    # Instantiating Processes
    my_process = MyProcess()
    reader = Reader()

    # Connecting Processes
    reader.ref_port.connect_var(my_process.var)

    # Running
[...]

PR #669, opened 2023-04-21
Fixed automatic append of py_ports and var_ports

tim-shea:
Added a check to prevent re-appending the same PyPort or VarPort that is already in a process model's py_ports or var_ports list.

Adjusted CspRecvPort implementation to allocate numpy _result array once at initialization rather than on every update.

Adjusted PyRefPortVectorDense to allocate header numpy arrays the first time they are used, rather than every time a value is sent.

Issue Number: #586 

Objective of pull request:
Fix incorrect append of py_ports and var_ports on each process model update, resulting in significant slowdown over time when using varports (e.g. Monitor).

- [Y] [Issue](https://github.com/lava-nc/lava/issues) created that explains the change and why it's needed
- [N] Tests are part of the PR (for bug fixes / features)
- [N] [Docs](https://github.com/lava-nc/docs) reviewed and added / updated if needed (for bug fixes / features)
- [Y] PR conforms to [Coding Conventions](https://lava-nc.org/developer_guide.html#coding-conventions)
- [Y] [PR applys BSD 3-clause or LGPL2.1+ Licenses](https://lava-nc.org/developer_guide.html#add-a-license) to all code files
- [Y] Lint (`flakeheaven lint src/lava tests/`) and (`bandit -r src/lava/.`) pass locally

One test in test_learning_rule.py seems to be stalling, I don't think this is due to these changes, but will investigate further.

Please check your PR type:
- [X] Bugfix

- Execution time of .read() on a RefPort grows the more we call it.

## What is the new behavior?
- Execution time of .read() on a RefPort is stable over the course of thousands of timesteps.

- [X] No

### DIFF
--- a/src/lava/magma/compiler/channels/pypychannel.py
+++ b/src/lava/magma/compiler/channels/pypychannel.py
@@ -238,7 +238,6 @@ class CspRecvPort(AbstractCspRecvPort):
             )
             for i in range(self._size)
         ]
-        self._result = np.ndarray(shape=self._shape, dtype=self._dtype)
         self._queue = CspRecvQueue(self._size)
         self.thread = Thread(
             target=self._req_callback,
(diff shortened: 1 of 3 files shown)


tim-shea:
Confirmed that this breaks some unit tests in test_learning_rule.py. Needs further investigation.

mgkwill (approved):

PR #675, opened 2023-04-27
Bugfix to enable monitoring aliased vars

AlessandroPierro:
Issue Number: #676 

Objective of pull request:

## Pull request checklist

## Pull request type

 

 

Please check your PR type:

- [x] Bugfix

- An assertion was raised when it was requested to create two different implicit var ports with the same name on the same process

## What is the new behavior?

- The different var ports will be created without assertion, with a numerical suffix added to the name

- [x] No

### DIFF
--- a/src/lava/magma/core/process/ports/ports.py
+++ b/src/lava/magma/core/process/ports/ports.py
@@ -646,16 +646,11 @@ class RefPort(AbstractRVPort, AbstractSrcPort):
             vp.process = var.process
             # VarPort name could shadow existing attribute
             if hasattr(var.process, vp.name):
-                found_unique_name = False
-                name = ""
-                i = 0
-                # Find a unique name
-                while not found_unique_name:
-                    i += 1
-                    name = vp.name + "_" + str(i)
-                    if not hasattr(var.process, name):
-                        found_unique_name = True
-                vp.name = name
+                name = str(vp.name)
+                name_suffix = 0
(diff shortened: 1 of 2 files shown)


AlessandroPierro:
Required to merge: https://github.com/lava-nc/lava-optimization/pull/217

PhilippPlank (approved):

src/lava/magma/core/process/ports/ports.py
```diff
@@ -646,9 +646,16 @@ def create_implicit_var_port(var: Var) -> "ImplicitVarPort":
             vp.process = var.process
             # VarPort name could shadow existing attribute
             if hasattr(var.process, vp.name):
-                raise AssertionError(
-                    "Name of implicit VarPort might conflict"
-                    " with existing attribute.")
+                found_unique_name = False
```
weidel-p, line 649: could we write that a bit more elegantly? 

Maybe something like:
```
name_suffix = 0
while hasattr(var.process, vp.name):
    vp.name += str(name_suffix)
    name_suffix += 1
```

weidel-p (changes requested):
This PR adds a suffix to a port name if it already exists which is an easy solution to the problem described in the issue.

I have one request though, could you please add a test which shows that the original issue is solved? So that we can probe aliased vars?

weidel-p (approved):

PR #686, opened 2023-05-12
First prototype of IO bridge Processes

gkarray:
Issue Number: 

Objective of pull request: Addition of IO bridge (Python only) Processes for getting input to/output from Lava.

## Pull request checklist

## Pull request type

 

 

Please check your PR type:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] Documentation changes

- At the moment, the only ways to get data into Lava, from a Python application, is to go through the `Dataloader` or `RingBuffer` Processes, or the `set()` method on Process `Var`.
  - The `Dataloader` Process keeps data in disk, and only loads portions of it step-by-step.
  - The `RingBuffer` Process loads all data to memory from the start.
  - The `set()` method requires running Lava workloads one time step at a time, and calling it between each `run()` call.
-  Similarly, the only ways to get data out of Lava, to use in a Python application, is to go through the `RingBuffer` Process, or the `get()` method on Process `Var`.
  - The `RingBuffer` Process buffers data in memory, and one has to call `get()` at the end of the run on its `data` `Var` to get the data out.
  - The `get()` method requires running Lava workloads one time step at a time, and calling it between each `run()` call.

## What is the new behavior?

- With the `InputBridge` Process on the input side, users would be able to seamlessly integrate Lava workloads (running in non-blocking mode) into broader applications. These Lava workloads would be able to get **dynamically generated input data from Python applications, without having to pause and re-run every time step.**
  - "Dynamically generated input data" means: Without having to pre-load the data from the start (in contrast with the `RingBuffer` option).
  - "From Python applications" means: Not loading from disk (in contrast with the `Dataloader` option). 
  - "Without having to pause and re-run every time step" means: Lava workload won't have to run one time step at a time (in contrast with the `set()` option).
- With the `OutputBridge` Process on the output side, users would be able to seamlessly integrate Lava workloads (running in non-blocking mode) into broader applications. These broader applications would be able to get **real-time output data from Lava workloads, without having to pause and re-run every time step.**
  - "Real-time output data from Lava workloads" means: Without having to wait for a Lava run to finish to get data out (in contrast with the `RingBuffer` option).
  -  "Without having to pause and re-run every time step" means: Lava workload won't have to run one time step at a time (in contrast with the `get()` option).

- [ ] Yes
- [x] No

### TODO:
- Add `RefPortInputBridge` variant, which relays data from Python to Lava `RefPort`.
- Add `VarOutputBridge` variant, which relays data from Lava `Var` to Python.
- Add variants with `VEC_SPARSE` `InPort`, `OutPort`, `RefPort`.
  - **These would have to be new Processes, not ProcessModels.** 
- __(Uncertain)__ Add (or change to) `AsyncProcessModels`, implementing the `AsyncProtocol`.

### Open questions:
- Do we really need `AsyncProcessModels` ? If so, do we need to change the current `PyLoihiProcessModels` to Async ones or do we need to simply add Async ones ?
- Should we create new fully fledged Lava constructs to implement this, or is the simple use of `multiprocessing.Pipe` object enough ? How would that behave, and how much would we need to change, if we switch to a different message passing back-end ?
  - At the moment, usage of `multiprocessing.Pipe` is restricted to piping objects with a limited size (which is, it seems, variable, from one machine/OS to another). For instance, as an example, we cannot send `np.ndarray` of `float` `dtype` with more than 2010 items.
  - Not only this, but even with objects of smaller size, we are limited by the number of different objects we can send in a row without receiving on the receiving end.
  - This is due to the fact that `multiprocessing.Pipe` construct has a limited buffer size.
  **- How should we address this ?**

### DIFF
--- a/src/lava/proc/io/input_bridge.py
+++ b/src/lava/proc/io/input_bridge.py
@@ -1,0 +1,80 @@
+import numpy as np
+import multiprocessing as mp
+
+from lava.magma.core.process.process import AbstractProcess
+from lava.magma.core.process.ports.ports import OutPort
+
+from lava.magma.core.resources import CPU
+from lava.magma.core.decorator import implements, requires, tag
+from lava.magma.core.model.py.model import PyLoihiProcessModel
+from lava.magma.core.sync.protocols.loihi_protocol import LoihiProtocol
+from lava.magma.core.model.py.type import LavaPyType
+from lava.magma.core.model.py.ports import PyOutPort
+
+
+# TODO : (GK) Is this really a good name ?
+class InputBridge(AbstractProcess):
+    """Process that gets data into Lava.
+
+    Parameters
+    ----------
+    out_shape : tuple
+        Shape of the OutPort of the Process.
+    """
+
+    def __init__(self, out_shape: tuple[int, ...]) -> None:
+        # TODO : (GK) I don't like having code before the call to super()... Is there a better way to do it ?
+        recv_pipe, self._send_pipe = mp.Pipe(duplex=False)
+        super().__init__(out_shape=out_shape, recv_pipe=recv_pipe)
+
+        self.out_port = OutPort(shape=out_shape)
+
+    def send_data(self, data: np.ndarray) -> None:
+        """Send data to the ProcessModel.
+
+        The data sent to the ProcessModel will be received by it, and then sent by the ProcessModel through the OutPort.
+
+        Parameters
+        ----------
+        data : np.ndarray
+            Data to get into Lava.
+        """
+        if data.shape != self.out_port.shape:
+            raise ValueError(f"Shape of data to send must be the same as the shape of the OutPort. "
+                             f"{data.shape=} != out_port.shape={self.out_port.shape}.")
+
+        self._send_pipe.send(data)
+
+
+# TODO : (GK) What's the best convention for naming ProcessModels ?
+@implements(proc=InputBridge, protocol=LoihiProtocol)
+@requires(CPU)
+@tag("floating_pt")
+class PyLoihiFloatingPointInputBridgeProcessModel(PyLoihiProcessModel):
+    """PyLoihiFloatingPointProcessModel for the InputBridge Process."""
+    out_port: PyOutPort = LavaPyType(PyOutPort.VEC_DENSE, float)
+
+    def __init__(self, proc_params: dict) -> None:
+        super().__init__(proc_params=proc_params)
+
+        self._recv_pipe: mp.connection.PipeConnection = self.proc_params["recv_pipe"]
+
+    def run_spk(self) -> None:
+        self.out_port.send(self._recv_pipe.recv())
+
+
+# TODO : (GK) What's the best convention for naming ProcessModels ?
+@implements(proc=InputBridge, protocol=LoihiProtocol)
+@requires(CPU)
+@tag("fixed_pt")
+class PyLoihiFixedPointInputBridgeProcessModel(PyLoihiProcessModel):
+    """PyLoihiFixedPointProcessModel for the InputBridge Process."""
+    out_port: PyOutPort = LavaPyType(PyOutPort.VEC_DENSE, int)
+
+    def __init__(self, proc_params: dict) -> None:
+        super().__init__(proc_params=proc_params)
+
+        self._recv_pipe: mp.connection.PipeConnection = self.proc_params["recv_pipe"]
+
+    def run_spk(self) -> None:
+        self.out_port.send(self._recv_pipe.recv())
--- a/src/lava/proc/io/output_bridge.py
+++ b/src/lava/proc/io/output_bridge.py
@@ -1,0 +1,76 @@
+import numpy as np
+import multiprocessing as mp
+
+from lava.magma.core.process.process import AbstractProcess
+from lava.magma.core.process.ports.ports import InPort
+
+from lava.magma.core.resources import CPU
+from lava.magma.core.decorator import implements, requires, tag
+from lava.magma.core.model.py.model import PyLoihiProcessModel
+from lava.magma.core.sync.protocols.loihi_protocol import LoihiProtocol
+from lava.magma.core.model.py.type import LavaPyType
+from lava.magma.core.model.py.ports import PyInPort
+
+
+# TODO : (GK) Is this really a good name ?
+class OutputBridge(AbstractProcess):
+    """Process that gets data out of Lava.
+
+    Parameters
+    ----------
+    shape_in : tuple
+        Shape of the InPort of the Process.
+    """
+
+    def __init__(self, in_shape: tuple[int, ...]) -> None:
+        # TODO : (GK) I don't like having code before the call to super()... Is there a better way to do it ?
+        self._recv_pipe, send_pipe = mp.Pipe(duplex=False)
+        super().__init__(in_shape=in_shape, send_pipe=send_pipe)
+
+        self.in_port = InPort(shape=in_shape)
+
+    def receive_data(self) -> np.ndarray:
+        """Receive data from the ProcessModel.
+
+        The data received from the InPort is sent from the ProcessModel and returned by this method.
+
+        Returns
+        ----------
+        data : np.ndarray
+            Data got from Lava.
+        """
+        return self._recv_pipe.recv()
+
+
+# TODO : (GK) What's the best convention for naming ProcessModels ?
+@implements(proc=OutputBridge, protocol=LoihiProtocol)
+@requires(CPU)
+@tag("floating_pt")
+class PyLoihiFloatingPointOutputBridgeProcessModel(PyLoihiProcessModel):
+    """PyLoihiFloatingPointProcessModel for the OutputBridge Process."""
+    in_port: PyInPort = LavaPyType(PyInPort.VEC_DENSE, float)
+
+    def __init__(self, proc_params: dict) -> None:
+        super().__init__(proc_params=proc_params)
+
+        self._send_pipe: mp.connection.PipeConnection = self.proc_params["send_pipe"]
+
+    def run_spk(self) -> None:
+        self._send_pipe.send(self.in_port.recv())
+
+
+# TODO : (GK) What's the best convention for naming ProcessModels ?
+@implements(proc=OutputBridge, protocol=LoihiProtocol)
+@requires(CPU)
+@tag("fixed_pt")
+class PyLoihiFixedPointOutputBridgeProcessModel(PyLoihiProcessModel):
+    """PyLoihiFixedPointProcessModel for the OutputBridge Process."""
+    in_port: PyInPort = LavaPyType(PyInPort.VEC_DENSE, float)
+
+    def __init__(self, proc_params: dict) -> None:
+        super().__init__(proc_params=proc_params)
+
+        self._send_pipe: mp.connection.PipeConnection = self.proc_params["send_pipe"]
+
+    def run_spk(self) -> None:
+        self._send_pipe.send(self.in_port.recv())
--- a/tests/lava/proc/io/test_input_bridge.py
+++ b/tests/lava/proc/io/test_input_bridge.py
@@ -1,0 +1,225 @@
+import numpy as np
+import unittest
+import multiprocessing as mp
+
+from lava.magma.core.process.process import AbstractProcess
+from lava.magma.core.process.ports.ports import InPort, OutPort
+from lava.magma.core.process.variable import Var
+
+from lava.magma.core.sync.protocols.loihi_protocol import LoihiProtocol
+
+from lava.magma.core.decorator import implements, requires, tag
+from lava.magma.core.resources import CPU
+
+from lava.magma.core.model.py.model import PyLoihiProcessModel
+from lava.magma.core.model.py.type import LavaPyType
+from lava.magma.core.model.py.ports import PyInPort, PyOutPort
+
+from lava.magma.core.run_configs import Loihi2SimCfg
+from lava.magma.core.run_conditions import RunSteps, RunContinuous
+
+from lava.proc.io.input_bridge import InputBridge, \
+    PyLoihiFloatingPointInputBridgeProcessModel, \
+    PyLoihiFixedPointInputBridgeProcessModel
+
+
+class Recv(AbstractProcess):
+    """Process that receives arbitrary dense data and stores it in a Var.
+
+    Parameters
+    ----------
+    shape: tuple
+        Shape of the InPort and Var.
+    """
+    def __init__(self, shape: tuple[...]):
+        super().__init__(shape=shape)
+
+        self.var = Var(shape=shape, init=0)
+        self.in_port = InPort(shape=shape)
+
+
+@implements(proc=Recv, protocol=LoihiProtocol)
+@requires(CPU)
+@tag("floating_pt")
+class PyLoihiFloatingPointRecvProcessModel(PyLoihiProcessModel):
+    """Receives dense floating point data from PyInPort and stores it in a Var."""
+    var: np.ndarray = LavaPyType(np.ndarray, float)
+    in_port: PyInPort = LavaPyType(PyInPort.VEC_DENSE, float)
+
+    def run_spk(self) -> None:
+        self.var = self.in_port.recv()
+
+
+@implements(proc=Recv, protocol=LoihiProtocol)
+@requires(CPU)
+@tag("fixed_pt")
+class PyLoihiFixedPointRecvProcessModel(PyLoihiProcessModel):
+    """Receives dense fixed point data from PyInPort and stores it in a Var."""
+    var: np.ndarray = LavaPyType(np.ndarray, int)
+    in_port: PyInPort = LavaPyType(PyInPort.VEC_DENSE, int)
+
+    def run_spk(self) -> None:
+        self.var = self.in_port.recv()
+
+
+class TestInputBridge(unittest.TestCase):
+    def test_init(self):
+        out_shape = (1, )
+
+        input_bridge = InputBridge(out_shape=out_shape)
+
+        self.assertIsInstance(input_bridge, InputBridge)
+        self.assertIsInstance(input_bridge._send_pipe, mp.connection.PipeConnection)
+        self.assertIsInstance(input_bridge.proc_params["recv_pipe"], mp.connection.PipeConnection)
+        self.assertEqual(input_bridge.out_port.shape, out_shape)
+
+    # TODO : Is this test really necessary ?
+    # def test_invalid_shape(self):
+    #     out_shape = (1.5, )
+    #
+    #     with self.assertRaises(ValueError):
+    #         InputBridge(out_shape=out_shape)
+
+    def test_send_data(self):
+        out_shape = (1,)
+
+        input_bridge = InputBridge(out_shape=out_shape)
+
+        send_data = np.ones(out_shape)
+        input_bridge.send_data(send_data)
+
+        recv_data = input_bridge.proc_params["recv_pipe"].recv()
+
+        np.testing.assert_equal(recv_data, send_data)
+
+    def test_send_data_shape_different_from_out_shape(self):
+        out_shape = (1,)
+
+        input_bridge = InputBridge(out_shape=out_shape)
+
+        data_shape = (1, 2)
+        send_data = np.ones(data_shape)
+
+        with self.assertRaises(ValueError):
+            input_bridge.send_data(send_data)
+
+    def test_send_data_multiple_times_before_receiving(self):
+        # limit for (1, ) = 54
+        # limit for (2, ) = 52
+        # limit for (3, ) = 51
+
+        # limit for (4, ) = 50
+        # limit for (4,1) = 49
+        # limit for (2,2) = 49
+
+        # limit for (50, ) = 23
+        # limit for (50, 1) = 23
+        # limit for (5, 10) = 23
+
+        # limit for (100, ) = 14
+        # limit for (500, ) = 3
+        # limit for (986, ) = 2
+        # limit for (2010, ) = 1
+        # limit for (2011, ) = 0
+
+        num_consecutive_send_before_receive = 50
+        out_shape = (4, )
+
+        input_bridge = InputBridge(out_shape=out_shape)
+
+        complete_send_data = [np.full(out_shape, i) for i in range(num_consecutive_send_before_receive)]
+
+        for send_data in complete_send_data:
+            input_bridge.send_data(send_data)
+
+        complete_recv_data = []
+
+        for _ in range(num_consecutive_send_before_receive):
+            complete_recv_data.append(input_bridge.proc_params["recv_pipe"].recv())
+
+        np.testing.assert_equal(complete_recv_data, complete_send_data)
+
+
+class TestInputBridgeFloatingPointProcessModel(unittest.TestCase):
+    def test_init(self):
+        out_shape = (1, )
+        recv_pipe, _ = mp.Pipe(duplex=False)
+        proc_params = {
+            "out_shape": out_shape,
+            "recv_pipe": recv_pipe
+        }
+
+        process_model = PyLoihiFloatingPointInputBridgeProcessModel(proc_params=proc_params)
+
+        self.assertIsInstance(process_model, PyLoihiFloatingPointInputBridgeProcessModel)
+        self.assertIsInstance(process_model._recv_pipe, mp.connection.PipeConnection)
+
+    def test_run(self):
+        num_steps = 1
+        data_shape = (1,)
+        send_data = np.ones(data_shape)
+
+        input_bridge = InputBridge(out_shape=data_shape)
+        recv = Recv(shape=data_shape)
+
+        input_bridge.out_port.connect(recv.in_port)
+
+        # run_condition has to be non-blocking in order to get past the call to run
+        run_condition = RunSteps(num_steps=num_steps, blocking=False)
+        run_cfg = Loihi2SimCfg(select_tag="floating_pt")
+
+        input_bridge.run(condition=run_condition, run_cfg=run_cfg)
+
+        input_bridge.send_data(send_data)
+
+        input_bridge.wait()
+
+        recv_data = recv.var.get()
+
+        input_bridge.stop()
+
+        np.testing.assert_equal(recv_data, send_data)
+
+
+class TestInputBridgeFixedPointProcessModel(unittest.TestCase):
+    def test_init(self):
+        out_shape = (1, )
+        recv_pipe, _ = mp.Pipe(duplex=False)
+        proc_params = {
+            "out_shape": out_shape,
+            "recv_pipe": recv_pipe
+        }
+
+        process_model = PyLoihiFixedPointInputBridgeProcessModel(proc_params=proc_params)
+
+        self.assertIsInstance(process_model, PyLoihiFixedPointInputBridgeProcessModel)
+        self.assertIsInstance(process_model._recv_pipe, mp.connection.PipeConnection)
+
+    def test_run(self):
+        num_steps = 1
+        data_shape = (1,)
+        send_data = np.ones(data_shape, dtype=int)
+
+        input_bridge = InputBridge(out_shape=data_shape)
+        recv = Recv(shape=data_shape)
+
+        input_bridge.out_port.connect(recv.in_port)
+
+        # run_condition has to be non-blocking in order to get past the call to run
+        run_condition = RunSteps(num_steps=num_steps, blocking=False)
+        run_cfg = Loihi2SimCfg(select_tag="fixed_pt")
+
+        input_bridge.run(condition=run_condition, run_cfg=run_cfg)
+
+        input_bridge.send_data(send_data)
+
+        input_bridge.wait()
+
+        recv_data = recv.var.get()
+
+        input_bridge.stop()
+
+        np.testing.assert_equal(recv_data, send_data)
+
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/lava/proc/io/test_output_bridge.py
+++ b/tests/lava/proc/io/test_output_bridge.py
@@ -1,0 +1,211 @@
+import numpy as np
+import unittest
+import multiprocessing as mp
+
+from lava.magma.core.process.process import AbstractProcess
+from lava.magma.core.process.ports.ports import InPort, OutPort
+from lava.magma.core.process.variable import Var
+
+from lava.magma.core.sync.protocols.loihi_protocol import LoihiProtocol
+
+from lava.magma.core.decorator import implements, requires, tag
+from lava.magma.core.resources import CPU
+
+from lava.magma.core.model.py.model import PyLoihiProcessModel
+from lava.magma.core.model.py.type import LavaPyType
+from lava.magma.core.model.py.ports import PyInPort, PyOutPort
+
+from lava.magma.core.run_configs import Loihi2SimCfg
+from lava.magma.core.run_conditions import RunSteps, RunContinuous
+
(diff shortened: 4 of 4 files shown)


src/lava/proc/io/input_bridge.py
```diff
@@ -0,0 +1,80 @@
...
+        Parameters
+        ----------
+        data : np.ndarray
+            Data to get into Lava.
+        """
+        if data.shape != self.out_port.shape:
+            raise ValueError(f"Shape of data to send must be the same as the shape of the OutPort. "
+                             f"{data.shape=} != out_port.shape={self.out_port.shape}.")
+
+        self._send_pipe.send(data)
```
tim-shea, line 46: This can block if the pipe is full. Not super likely, but it can happen if a data generating process runs very quickly while this consumer runs much more slowly.

src/lava/proc/io/input_bridge.py
```diff
@@ -0,0 +1,80 @@
...
+    """PyLoihiFloatingPointProcessModel for the InputBridge Process."""
+    out_port: PyOutPort = LavaPyType(PyOutPort.VEC_DENSE, float)
+
+    def __init__(self, proc_params: dict) -> None:
+        super().__init__(proc_params=proc_params)
+
+        self._recv_pipe: mp.connection.PipeConnection = self.proc_params["recv_pipe"]
+
+    def run_spk(self) -> None:
+        self.out_port.send(self._recv_pipe.recv())
```
tim-shea, line 63: recv will block if the pipe is empty, and raise an EOFError if the pipe is closed on the other end. Both should be documented/handled appropriately.

src/lava/proc/io/input_bridge.py
```diff
@@ -0,0 +1,80 @@
...
+            Data to get into Lava.
+        """
+        if data.shape != self.out_port.shape:
+            raise ValueError(f"Shape of data to send must be the same as the shape of the OutPort. "
+                             f"{data.shape=} != out_port.shape={self.out_port.shape}.")
+
+        self._send_pipe.send(data)
+
+
+# TODO : (GK) What's the best convention for naming ProcessModels ?
```
tim-shea, line 49: Unfortunately we never made any further progress on how to reduce the combinatorial naming explosion.

Nobody should really even see these types of tagged PMs, so whatever you name them is probably fine.

src/lava/proc/io/output_bridge.py
```diff
@@ -0,0 +1,76 @@
...
+from lava.magma.core.resources import CPU
+from lava.magma.core.decorator import implements, requires, tag
+from lava.magma.core.model.py.model import PyLoihiProcessModel
+from lava.magma.core.sync.protocols.loihi_protocol import LoihiProtocol
+from lava.magma.core.model.py.type import LavaPyType
+from lava.magma.core.model.py.ports import PyInPort
+
+
+# TODO : (GK) Is this really a good name ?
+class OutputBridge(AbstractProcess):
```
tim-shea, line 16: Same comments as above.

tim-shea:
Since I don't see any replies but I do see a bunch of new commits, I'm not sure what you think of my comments above, but here's a more concrete proposal for naming:

input_bridge.py > input_synchronizer.py
InputBridge > InputSynchronizer
AbstractPyLoihiInputBridgeProcessModel > PyInputSynchronizerModel
PyLoihiFloatingPointInputBridgeProcessModel > PyInputSynchronizerModelFloat
PyLoihiFixedPointInputBridgeProcessModel > PyInputSynchronizerModelFixed

output_bridge.py > output_synchronizer.py
OutputBridge > OutputSynchronizer
AbstractPyLoihiOutputBridgeProcessModel > PyOutputSynchronizerModel
etc...

You really don't need async protocol models for the synchronizer processes, because the way they're written makes them useful specifically for synchronizing your code to your synchronous Loihi model, hence the renames to describe their actual function.

See also Lif models.py, Dense models.py, etc, where the models start with Py, not PyLoihi, and include Model, but not ProcessModel.

Continuing:
in_bridge.py > async_injector.py
AsyncInputBridge > AsyncInjector
AsyncProcessDenseModel > AsyncInjectorModelFloat

Note that the process name should describe the behavior of all models, in this case Async should not refer to Async protocol vs Loihi protocol, but to the behavior in which the input is injected asynchronously with respect to the updates of the connected port. Basically, this process should allow me to sporadically send data or recv data without caring whether I send or recv the correct number of times, and without ever blocking my calling code or the connected port.

out_bridge.py > async_extractor.py
AsyncOutputBridge > AsyncExtractor
AsyncProcessModel > AsyncExtractorModelFloat

PhilippPlank (approved):

src/lava/proc/io/input_bridge.py
```diff
@@ -0,0 +1,80 @@
...
+from lava.magma.core.resources import CPU
+from lava.magma.core.decorator import implements, requires, tag
+from lava.magma.core.model.py.model import PyLoihiProcessModel
+from lava.magma.core.sync.protocols.loihi_protocol import LoihiProtocol
+from lava.magma.core.model.py.type import LavaPyType
+from lava.magma.core.model.py.ports import PyOutPort
+
+
+# TODO : (GK) Is this really a good name ?
+class InputBridge(AbstractProcess):
```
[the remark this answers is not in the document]
gkarray: Changed the name of the Process to Injector.

src/lava/proc/io/input_bridge.py
gkarray: There is now an a parameter permitting the user to choose whether or not to block when the channel is full.

src/lava/proc/io/input_bridge.py
gkarray: There is now an a parameter permitting the user to choose whether or not to block when the channel is full.
As for the EOFError, it is not raised with the new implementation.

src/lava/proc/io/bridge/extractor.py
```diff
@@ -0,0 +1,179 @@
...
+from lava.proc.io.bridge.utils import ChannelConfig, ChannelSendBufferFull, \
+    ChannelRecvBufferEmpty, ChannelRecvBufferNotEmpty,\
+    validate_shape, validate_buffer_size, validate_channel_config, \
+    send_data_blocking, send_data_non_blocking_drop, \
+    recv_empty_blocking, recv_empty_non_blocking_zeros, recv_not_empty_fifo, \
+    recv_not_empty_accumulate
+
+
+class Extractor(AbstractProcess):
+    """Process that gets data out of Lava.
```
tim-shea, line 27: You can cut this first line of the docstring and the resulting text will be shorter and more accurate.

src/lava/proc/io/bridge/extractor.py
```diff
@@ -0,0 +1,179 @@
...
+
+    The Extractor Process exposes a method, recv_data, which enables users to
+    receive data in external Python scripts from the ProcessModel of the
```
tim-shea, line 30: Think about this from the perspective of What problem does this process solve?

E.g. "Extractor allows non-Lava code, such as a third-party Python library or an actuator, to extract data from a synchronous Lava process by calling recv_data once per timestep."

src/lava/proc/io/bridge/extractor.py
```diff
@@ -0,0 +1,179 @@
...
+    Process, while the Lava Runtime is running.
+    Internally, this Process builds a channel (extractor_channel, of type
+    PyPyChannel).
+    The src_port of the channel lives in the ProcessModel.
+    The dst_port of the channel lives in the Process.
+
+    In the ProcessModel, data is received from this Process's InPort,
```
tim-shea, line 37: The rest of this docstring is probably longer than needed. Just summarize what it does, not exactly how it does it.

src/lava/proc/io/bridge/extractor.py
```diff
@@ -0,0 +1,179 @@
...
+        validate_shape(shape)
+        validate_buffer_size(buffer_size)
+        validate_channel_config(extractor_channel_config)
+
+        self._shape = shape
+
+        self._extractor_channel_config = extractor_channel_config
+
+        self._multi_processing = MultiProcessing()
+        self._multi_processing.start()
```
tim-shea, line 87: This is not ideal. MultiProcessing is meant to be shared by all the Processes which use the same Runtime to ensure that all shared memory is cleaned up consistently.

Instead of doing this, you could defer creation of the PyPyChannel until the runtime is set for this process (`process.run` calls `p.runtime = runtime for p in all_connected_processes`). Once that's done, you can get the runtime._messaging_infrastructure, which will be an instance of MultiProcessing, and use that instead of creating your own instance per Injector/Extractor.

src/lava/proc/io/bridge/extractor.py
```diff
@@ -0,0 +1,179 @@
...
+
+        self._extractor_channel = \
+            PyPyChannel(message_infrastructure=self._multi_processing,
+                        src_name="src",
+                        dst_name="dst",
+                        shape=self._shape,
+                        dtype=float,
+                        size=buffer_size)
+        self._extractor_channel_dst_port = self._extractor_channel.dst_port
+        self._extractor_channel_dst_port.start()
```
tim-shea, line 97: I'm not positive, but I don't think you should be calling start on the channel until after the runtime has been initialized.

src/lava/proc/io/bridge/extractor.py
```diff
@@ -0,0 +1,179 @@
...
+        if self._extractor_channel_config.recv_buffer_not_empty == \
+                ChannelRecvBufferNotEmpty.FIFO:
+            self._recv_not_empty = recv_not_empty_fifo
+        elif self._extractor_channel_config.recv_buffer_not_empty == \
+                ChannelRecvBufferNotEmpty.ACCUMULATE:
+            self._recv_not_empty = recv_not_empty_accumulate
+
+        self.in_port = InPort(shape=shape)
+
+    def recv_data(self) -> np.ndarray:
```
tim-shea, line 120: I'm not a fan of gratuitous name shortening. Good, public APIs should prioritize clarity first, then conciseness.

Data is redundant, since everything that is ever sent or received in a computer is data, so a better name would be just receive.

src/lava/proc/io/bridge/extractor.py
```diff
@@ -0,0 +1,179 @@
...
+        else:
+            data = self._recv_not_empty(
+                dst_port=self._extractor_channel_dst_port,
+                zeros=np.zeros(self._shape),
+                elements_in_buffer=elements_in_buffer)
+
+        return data
+
+    def __del__(self) -> None:
+        super().__del__()
```
tim-shea, line 146: If you follow the comment under __init__, you should be able to get rid of this, which is good because this will leak whenever a notebook is killed/restarted.

src/lava/proc/io/bridge/injector.py
```diff
@@ -0,0 +1,176 @@
...
+from lava.proc.io.bridge.utils import ChannelConfig, ChannelSendBufferFull, \
+    ChannelRecvBufferEmpty, ChannelRecvBufferNotEmpty,\
+    validate_shape, validate_buffer_size, validate_channel_config, \
+    validate_send_data, send_data_blocking, send_data_non_blocking_drop, \
+    recv_empty_blocking, recv_empty_non_blocking_zeros, recv_not_empty_fifo, \
+    recv_not_empty_accumulate
+
+
+class Injector(AbstractProcess):
+    """Process that gets data into Lava.
```
tim-shea, line 26: All the comments for Extractor above also apply here.

src/lava/proc/io/bridge/extractor.py
```diff
@@ -0,0 +1,179 @@
...
+        data : np.ndarray
+            Data received.
+        """
+        elements_in_buffer = \
+            self._extractor_channel_dst_port._queue._qsize()
+
+        if elements_in_buffer == 0:
+            data = self._recv_empty(
+                dst_port=self._extractor_channel_dst_port,
+                zeros=np.zeros(self._shape))
```
tim-shea, line 136: Please try not to allocate memory inside runtime code. You should create this zeros array in __init__, then call fill(0) before reusing it if needed.

src/lava/proc/io/bridge/extractor.py
```diff
@@ -0,0 +1,179 @@
...
+        else:
+            data = self._recv_not_empty(
+                dst_port=self._extractor_channel_dst_port,
+                zeros=np.zeros(self._shape),
```
tim-shea, line 140: Same as above re memory allocation.

src/lava/proc/io/bridge/injector.py
```diff
@@ -0,0 +1,176 @@
...
+        """Send data to the ProcessModel.
+
+        The data is sent through injector_channel.src_port.
+
+        Parameters
+        ----------
+        data : np.ndarray
+            Data to be sent.
+        """
+        validate_send_data(data, self.out_port)
```
tim-shea, line 121: For performance critical code, it's best not to put any manual type checking / parameter assertions in the "tight loop".

In this case, it's better to let the incorrect data go through and generate a probably worse error message later, than to pay a performance cost on every single timestep of a correctly implemented model. This is especially true here because static type checking can be used in any modern IDE to let people know they're passing the wrong kind of thing.

src/lava/proc/io/bridge/utils.py
```diff
@@ -0,0 +1,275 @@
...
+    ----------
+    zeros : np.ndarray
+        Zeros array to be returned.
+
+    Returns
+    ----------
+    data : np.ndarray
+        Data received (zeros).
+    """
+    return zeros
```
tim-shea, line 226: Is this not implemented, or is this a dummy method to satisfy a consistent interface to other recv methods?

tim-shea (changes requested):
Very nice! Great cleanups, this is a clear, simple, and super useful little addition to the core Lava API.

Only top level suggestion to add is to drop the "bridge" in module paths, and just locate these three modules in io.

src/lava/proc/io/bridge/utils.py
gkarray: This is a dummy method to satisfy a consistent interface to other recv methods.

src/lava/proc/io/bridge/extractor.py
gkarray: Changed this.

src/lava/proc/io/bridge/extractor.py
gkarray: Changed this.

src/lava/proc/io/bridge/extractor.py
gkarray: Changed this to receive, and send_data in the Injector to send.

src/lava/proc/io/bridge/extractor.py
gkarray: Made the docstrings shorter.

However, I left a small part explaining how it is doing what it is doing for clarity. I could remove it as well if you still think it is not necessary.

src/lava/proc/io/bridge/extractor.py
gkarray: Removed this line.

src/lava/proc/io/utils.py
```diff
@@ -0,0 +1,255 @@
+from enum import IntEnum, auto
```
mathisrichter, line 1: Missing license header

tests/lava/proc/io/test_injector.py
```diff
@@ -0,0 +1,576 @@
...
+            injector._injector_channel_config.send_buffer_full,
+            ChannelSendBufferFull.BLOCKING)
+        self.assertEqual(
+            injector._injector_channel_config.recv_buffer_empty,
+            ChannelRecvBufferEmpty.BLOCKING)
+        self.assertEqual(
+            injector._injector_channel_config.recv_buffer_not_empty,
+            ChannelRecvBufferNotEmpty.FIFO)
+
+        self.assertIsInstance(injector._multi_processing, MultiProcessing)
```
mathisrichter, line 91: No testing of private members

mathisrichter (changes requested):
Looks good overall - some minor changes and naming suggestions.
We did this in a live review, so my comments are a bit short; more reminders for @gkarray .

tests/lava/proc/io/test_injector.py
gkarray: Resolved.

tests/lava/proc/io/test_injector.py
```diff
@@ -0,0 +1,576 @@
...
+import unittest
+import threading
+import time
+from queue import Queue
+import typing as ty
+
+from lava.magma.core.process.process import AbstractProcess
+from lava.magma.core.process.ports.ports import InPort, OutPort
+from lava.magma.core.process.variable import Var
+
```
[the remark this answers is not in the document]
gkarray: Resolved

src/lava/proc/io/utils.py
gkarray: Resolved.

src/lava/proc/io/injector.py
```diff
@@ -0,0 +1,159 @@
+import numpy as np
```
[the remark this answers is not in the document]
gkarray: Resolved.

src/lava/proc/io/extractor.py
```diff
@@ -0,0 +1,161 @@
...
+        super().__init__(proc_params=proc_params)
+
+        self._shape = self.proc_params["shape"]
+        self._extractor_channel_config = \
+            self.proc_params["extractor_channel_config"]
+        self._extractor_channel_src_port = self.proc_params[
+            "extractor_channel_src_port"]
+        self._extractor_channel_src_port.start()
+
+        if self._extractor_channel_config.send_buffer_full == \
```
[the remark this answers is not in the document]
gkarray: Resolved.

src/lava/proc/io/utils.py
```diff
@@ -0,0 +1,255 @@
...
+
+    Returns
+    ----------
+    data : np.ndarray
+        Data received.
+    """
+    return dst_port.recv()
+
+
+def recv_empty_non_blocking_zeros(zeros: np.ndarray, **kwargs) \
```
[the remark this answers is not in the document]
gkarray: Resolved.

src/lava/proc/io/utils.py
```diff
@@ -0,0 +1,255 @@
...
+    data : np.ndarray
+        Data to be sent.
+    """
+    if src_port.probe():
+        src_port.send(data)
+    else:
+        warnings.warn("Send buffer is full. Dropping items ...")
+
+
+def recv_empty_blocking(dst_port: CspRecvPort, **kwargs) -> np.ndarray:
```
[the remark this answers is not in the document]
gkarray: Resolved.

src/lava/proc/io/extractor.py
```diff
@@ -0,0 +1,161 @@
...
+                        size=buffer_size)
+        self._extractor_channel_dst_port = self._extractor_channel.dst_port
+        self._extractor_channel_dst_port.start()
+
+        self.proc_params["shape"] = shape
+        self.proc_params["extractor_channel_config"] = extractor_channel_config
+        self.proc_params["extractor_channel_src_port"] = \
+            self._extractor_channel.src_port
+
+        if self._extractor_channel_config.recv_buffer_empty == \
```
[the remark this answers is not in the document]
gkarray: Changed the names. Do you think they look better now ?

src/lava/proc/io/extractor.py
```diff
@@ -0,0 +1,161 @@
...
+from lava.magma.core.model.py.model import PyLoihiProcessModel
+from lava.magma.core.sync.protocols.loihi_protocol import LoihiProtocol
+from lava.magma.core.model.py.type import LavaPyType
+from lava.magma.core.model.py.ports import PyInPort
+
+from lava.magma.compiler.channels.pypychannel import PyPyChannel
+from lava.magma.runtime.message_infrastructure.multiprocessing import \
+    MultiProcessing
+
+from lava.proc.io.utils import ChannelConfig, ChannelSendBufferFull, \
```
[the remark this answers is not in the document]
gkarray: Resolved.

src/lava/proc/io/extractor.py
```diff
@@ -0,0 +1,161 @@
...
+        self._multi_processing.start()
+
+        self._extractor_channel = \
+            PyPyChannel(message_infrastructure=self._multi_processing,
+                        src_name="src",
+                        dst_name="dst",
+                        shape=self._shape,
+                        dtype=float,
+                        size=buffer_size)
+        self._extractor_channel_dst_port = self._extractor_channel.dst_port
```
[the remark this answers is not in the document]
gkarray: Resolved.

src/lava/proc/io/extractor.py
```diff
@@ -0,0 +1,161 @@
...
+                 extractor_channel_config: ty.Optional[ChannelConfig] =
+                 ChannelConfig()) -> None:
+        super().__init__()
+
+        validate_shape(shape)
+        validate_buffer_size(buffer_size)
+        validate_channel_config(extractor_channel_config)
+
+        self._shape = shape
+        self._extractor_channel_config = extractor_channel_config
```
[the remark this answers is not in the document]
gkarray: Changed the names. Do you think they look better now ?

src/lava/proc/io/extractor.py
```diff
@@ -0,0 +1,161 @@
...
+        Buffer size of the extractor_channel.
+    extractor_channel_config : optional, ChannelConfig
+        Configuration object specifying how the src_port behaves when the
+        buffer is full and how the dst_port behaves when the buffer is empty
+        and not empty.
+    """
+    def __init__(self,
+                 shape: tuple[int, ...],
+                 buffer_size: ty.Optional[int] = 50,
+                 extractor_channel_config: ty.Optional[ChannelConfig] =
```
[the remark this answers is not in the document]
gkarray: Resolved.

src/lava/proc/io/extractor.py
```diff
@@ -0,0 +1,161 @@
...
+    When the receive method is called from the external Python script, data is
+    received from the extractor_channel.dst_port.
+
+    Parameters
+    ----------
+    shape : tuple
+        Shape of the InPort of the Process.
+    buffer_size : optional, int
+        Buffer size of the extractor_channel.
+    extractor_channel_config : optional, ChannelConfig
```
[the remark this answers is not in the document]
gkarray: Resolved.

src/lava/proc/io/extractor.py
```diff
@@ -0,0 +1,161 @@
+import numpy as np
```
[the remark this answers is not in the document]
gkarray: Resolved.

src/lava/proc/io/bridge/extractor.py
```diff
@@ -0,0 +1,179 @@
+import numpy as np
```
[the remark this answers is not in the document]
gkarray: Resolved.

src/lava/proc/io/injector.py
```diff
@@ -0,0 +1,139 @@
...
+from lava.magma.core.model.py.type import LavaPyType
+from lava.magma.core.model.py.ports import PyOutPort
+from lava.magma.runtime.message_infrastructure.multiprocessing import \
+    MultiProcessing
+from lava.magma.compiler.channels.pypychannel import PyPyChannel
+from lava.proc.io import utils
+
+
+class Injector(AbstractProcess):
+    """Injector allows non-Lava code, such as a third-party Python library or
```
mathisrichter, line 23: Python library ~~or~~

src/lava/proc/io/injector.py
```diff
@@ -0,0 +1,139 @@
...
+
+    Parameters
+    ----------
+    shape : tuple
+        Shape of the OutPort of the Process, and of the np.ndarrays passed
+        through the channel between the Process and the ProcessModel.
+    buffer_size : int, optional
+        Buffer size (in terms of number of np.ndarrays) of the channel between
+        the Process and ProcessModel.
+    channel_config : optional, ChannelConfig
```
mathisrichter, line 45:
```python
channel_config : ChannelConfig, optional
```

src/lava/proc/io/utils.py
```diff
@@ -0,0 +1,272 @@
+# Copyright (C) 2021-22 Intel Corporation
```
mathisrichter, line 1: Copyright 2023

tests/lava/proc/io/test_extractor.py
```diff
@@ -0,0 +1,523 @@
+# Copyright (C) 2021-22 Intel Corporation
```
mathisrichter, line 1: Copyright 2023

tests/lava/proc/io/test_injector.py
```diff
@@ -0,0 +1,537 @@
+# Copyright (C) 2021-22 Intel Corporation
```
mathisrichter, line 1: Copyright 2023

tests/lava/proc/io/test_injector.py
```diff
@@ -0,0 +1,537 @@
...
+
+        injector.pause()
+        injector.wait()
+
+        recv_var_data = recv.var.get()
+
+        injector.stop()
+
+        np.testing.assert_equal(recv_var_data[:num_send // 10],
+                                send_data[:num_send // 10])
```
mathisrichter, line 537: The `:num_send // 10` needs to be changed to compare all sent vs received data or it needs to be documented.

tests/lava/proc/io/test_injector.py
mathisrichter, line 1: Please document what in the individual tests is the most important thing. You said that this is the order in which certain events happen. Please clarify this in the docstring of each test where this applies, also for the extractor.

mathisrichter (changes requested):

src/lava/proc/io/bridge/extractor.py
gkarray: I also do not like the fact of using another instance of MultiProcessing here. However, I encountered several hurdles when I tried to make use of the same instance as the one the Runtime uses.

1. If we want to make use of the same MultiProcessing instance as the Runtime, we need to somehow defer the creation of the PyPyChannel to the somewhere between the calls to `_build_message_passing_infrastructure` and `_build_processes` in the `initialize` method of Runtime. This is because the PyPyChannel needs a MultiProcessing instance to be instantiated (and the instance of MultiProcessing is created in `_build_message_passing_infrastructure`) and because the `dst_port` of the created PyPyChannel needs to be passed to the ProcModel through the `proc_params` (and the ProcModel is instantiated with the `proc_params` in `_build_processes`).
   - This is problematic because we have no way of triggering computation to happen in the Process at that time (remember, when the PyPyChannel is created, the `dst_port` needs to be passed to the ProcModel through the `proc_params`, but the Process needs to hold the `src_port` as well!)
   - We could circumvent this by making changes to the Runtime to actually have the possiblity of triggering some callback functions between `_build_message_passing_infrastructure` and `_build_processes` in the `initialize`, but this feels too specific to be added to Runtime.
2. A somehow very similar problem, but yet, arising in a slightly different case, is the following: Let's say you have an Injector->LIF network. And you call run() on the LIF Process, not on the Injector. It's true that internally, a call to run() on the LIF will trigger compilation, then creation of a Runtime, then it will assign the Runtime to all Processes. But in this case, it also needs to trigger some callback in the Injector Process to actually create the PyPyChannel etc...
   - The solution proposed for the first problem could solve this issue... But again, it is very much specific to the Injector and Extractor Processes to be added to the Runtime code.

src/lava/proc/io/bridge/extractor.py
gkarray: See comment under init.

src/lava/proc/io/bridge/extractor.py
gkarray: I also wanted to implement it this way. However, if we do that, and use the Injector or Extractor as we do in the `test_run_steps_non_blocking` tests (found in both `test_injector.py` and `test_extractor.py`) (which is how these Processes are typically expected to be used) will probably not work most of the times.
This is because the call to run(), when using the RunSteps(blocking=False) condition will not block, and by the time the next line executes (i.e we start sending or receiving), Runtime might not be initialized yet, thus an error would be thrown.
We could of course make sure to sleep a bit after the call to run() to give some time for the compilation to finish and for the Runtime to be initialized, but this amount of time needed for this could very much vary depending on the network being run...

src/lava/proc/io/extractor.py
```diff
@@ -0,0 +1,138 @@
+# Copyright (C) 2021-22 Intel Corporation
```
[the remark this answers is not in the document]
gkarray: Resolved

src/lava/proc/io/injector.py
gkarray: Resolved

src/lava/proc/io/utils.py
gkarray: Resolved

tests/lava/proc/io/test_extractor.py
gkarray: Resolved

tests/lava/proc/io/test_injector.py
gkarray: Resolved

tests/lava/proc/io/test_injector.py
gkarray: In the Injector, I added a comment explaining why this was happening.
In the Extractor, I got rid of it because it wasn't actually needed (i.e it now compares to all sent data).

tests/lava/proc/io/test_injector.py
gkarray: Refactored some of the tests to reduce duplication, and added explanation about the most important things in the relevant places.

mathisrichter (approved):

src/lava/proc/io/bridge/extractor.py
tim-shea: The runtime initialization should never be threaded/subprocessed. When you call proc.run, if you haven't already called proc.create_runtime, it will always block at least long enough to setup all the messaging and runtime infrastructure, create ports and channels, etc. The only thing that should be non-blocking is the work that happens inside the process builders and the run code itself.

Now, you might still be right that it could lead to errors if implemented that way, but those errors would be bugs that should be fixed, not just acceptable indeterminism.

gkarray: You are right. The call to run() with RunSteps(blocking=False) actually returns after the Runtime is initialized (it actually even returns after the Runtime starts running...), so the issue I was pointing to is not really an issue here.

That being said, starting the channel only after the Runtime is initialized requires us to be able to trigger some callback in the Process after the Runtime is initialized, which is not straightforward (see comment under init).

src/lava/proc/io/bridge/extractor.py
tim-shea: This all makes sense, yes, but there should be ways to work around it, and in my view such a workaround or solution is necessary to avoid the memory issues.

On looking into Runtime._build_channels, it seems like one option would be to use the default channel build mechanism. If you create both an InPort and an OutPort in the proc and connect them together, Runtime will create (and destroy) your PyPyChannel for you, and the ports will be accessible to the proc_model of course because they already get passed from the channel builder (which runs in the user thread) to the subprocess created by the process_builder. So, that only leaves that the process needs to be able to grab its end of the channel, which you should be able to do from Runtime either by searching inside `Runtime._open_ports` (though that might be tricky) or by grabbing the process_builder with `Runtime._get_process_builder_for_process(proc)` then `proc_builder.get_csp_ports()`

src/lava/proc/io/bridge/extractor.py
tim-shea: Resolved. We will merge working version and fix issues with MultiProcessing separately.

tim-shea (approved):
Great work Ghassen.

Issue #768, opened 2023-08-05
Process.run and Runtime.start should accept an int or None as conditions


tim-shea:
### User story
As a user, I want to be able to pass an int or None to the runtime.start method and process.run, rather than an AbstractRunCondition. If I pass an int, runtime.start should create a RunSteps condition with the value of the int as the number of steps. If I pass None, runtime.start should create a RunContinuous condition. This will reduce the need to import or know about the RunCondition classes for the vast majority of use cases.

### Conditions of satisfaction
- User can pass an AbstractRunCondition, int, or None for the condition parameter of Process.run
- User can pass an AbstractRunCondition, int, or None for the condition parameter of Runtime.start
- If the condition is an int, that will be used as the number of steps to run the process.
- If the condition is None, the process will run with a RunContinuous condition.

kds300:
Hi @tim-shea , I noticed this issue and I've actually been working on something similar. Rather than passing `int` or `None` to `process.run()`, I created a new class which takes `num_steps` and `blocking` as parameters. It uses the `__new__` method to then either create a `RunSteps` or `RunContinuous` instance depending on the type of `num_steps`.

I saw the PR you linked was closed, but I'd be happy to make a branch on my lava fork with the class I wrote. I haven't actually used it to call a `RunContinuous` instance (I never use that class), but I wrote some tests to make sure it creates an instance of the correct class.

PR #769, opened 2023-08-05
Enable int or None as run_condition for Process.run and Runtime.start

tim-shea:
Issue Number: #768 

Objective of pull request: Add logic to create a RunSteps or RunContinuous in Runtime.start when user passes an int or None respectively. This allows users to just call process.run(5) or process.run() and get useful behavior, rather than always importing RunSteps/RunContinuous.

## Pull request checklist
-   [X] [Issue](https://github.com/lava-nc/lava/issues) created that explains the change and why it's needed
-   [X] Tests are part of the PR (for bug fixes / features)
-   [X] [Docs](https://github.com/lava-nc/docs) reviewed and added / updated if needed (for bug fixes / features)
-   [X] PR conforms to [Coding Conventions](https://lava-nc.org/developer_guide.html#coding-conventions)
-   [X] [PR applys BSD 3-clause or LGPL2.1+ Licenses](https://lava-nc.org/developer_guide.html#add-a-license) to all code files
-   [X] Lint (`flakeheaven lint src/lava tests/`) and (`bandit -r src/lava/.`) pass locally
-   [X] Build tests (`pytest`) passes locally

## Pull request type
Please check your PR type:
-   [X] Feature

- User must always explicitly create an instance of AbstractRunCondition.

## What is the new behavior?
- Runtime.start will implicitly create a RunSteps or RunContinuous if condition is an int or None.

-   [X] No

### DIFF
--- a/src/lava/magma/core/process/process.py
+++ b/src/lava/magma/core/process/process.py
@@ -304,7 +304,7 @@ class AbstractProcess(metaclass=ProcessPostInitCaller):
             return False
 
     def run(self,
-            condition: AbstractRunCondition,
+            condition: ty.Optional[ty.Union[AbstractRunCondition, int]],
             run_cfg: ty.Optional[RunConfig] = None,
             compile_config: ty.Optional[ty.Dict[str, ty.Any]] = None):
         """Executes this and any connected Processes that form a Process
@@ -333,8 +333,10 @@ class AbstractProcess(metaclass=ProcessPostInitCaller):
 
         Parameters
         ----------
-        condition : AbstractRunCondition
-            Specifies for how long to run the Process.
+        condition : Optional[Union[AbstractRunCondition, int]], default=None
+            The condition that determines how long to run the process. If
+            condition is an int, runs for the specified number of timesteps.
+            If condition is None, runs until paused or stopped.
         run_cfg : RunConfig, optional
             Used by the compiler to select a ProcessModel for each Process.
             Must be provided when Processes have to be compiled, can be
--- a/src/lava/magma/runtime/runtime.py
+++ b/src/lava/magma/runtime/runtime.py
@@ -327,16 +327,27 @@ class Runtime:
                     self._is_running = False
                     return
 
-    def start(self, run_condition: AbstractRunCondition):
+    def start(self, run_condition:
+              ty.Optional[ty.Union[AbstractRunCondition, int]] = None):
         """
-        Given a run condition, starts the runtime
-
-        :param run_condition: AbstractRunCondition
-        :return: None
+        Start the runtime and run as long as the run_condition dictates.
+        Parameters
+        ----------
+        run_condition : Optional[Union[AbstractRunCondition, int]],
+            default = None
+            The condition specifying how long to run the runtime. If
+            run_condition is an int, the runtime will create a RunSteps
+            condition and run for the number of timesteps. If
+            run_condition is None (default), runtime will create a
+            RunContinuous condition and run until interrupted.
         """
         if self._is_initialized:
             # Start running
             self._is_started = True
+            if run_condition is None:
+                run_condition = RunContinuous()
+            elif isinstance(run_condition, int):
+                run_condition = RunSteps(run_condition)
             self._run(run_condition)
         else:
             self.log.info("Runtime not initialized yet.")
--- a/tests/lava/magma/runtime/test_runtime.py
+++ b/tests/lava/magma/runtime/test_runtime.py
@@ -21,7 +21,7 @@ from lava.proc.io.sink import RingBuffer as SinkBuffer
 
 
 def setup_condition_test():
-    source = SourceBuffer(data=np.array(range(1000)).reshape((1,-1)))
+    source = SourceBuffer(data=np.array(range(1000)).reshape((1, -1)))
     sink = SinkBuffer(shape=(1,), buffer=10)
     source.s_out.connect(sink.a_in)
     source.create_runtime(run_cfg=Loihi2SimCfg())
--- a/tests/lava/magma/runtime/test_runtime.py
+++ b/tests/lava/magma/runtime/test_runtime.py
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # See: https://spdx.org/licenses/
 
+import numpy as np
+import time
 import typing as ty
 import unittest
 from unittest.mock import Mock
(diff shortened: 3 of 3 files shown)


tests/lava/magma/runtime/test_runtime.py
```diff
@@ -73,6 +88,33 @@ def test_executable_node_config_assertion(self):
                          "Expected node_configs[0] node_config length to be 2")
         runtime4.stop()
 
+    def test_run_conditions(self):
```
mathisrichter, line 91: You are testing four features in this one test. I would split these into separate tests, where you reuse the setup with the `setUp()` method:
```python
def setUp() -> None:
    ...
    self.runtime = source.runtime
    ...

def test_implicit_run_steps_condition() -> None:
    self.runtime.start(run_condition=3)
    data = sink.data.get().astype(int).flatten().tolist()
    self.assertEqual(data, [0, 1, 2, 3, 4, 5, 6, 7, 8, 0])

def test_implicit_run_continuous_condition() -> None:
    self.runtime.start()
    time.sleep(0.02)
    self.runtime.pause()
    data = sink.data.get().astype(int).flatten().tolist()
    self.assertGreater(max(data), timestep + 10)
```

It may also be nice to add a little method for readability - just a thought:
```python
def get_data(process: AbstractProcess) -> ty.List[int]:
    return process.data.get().astype(int).flatten().tolist()

def test_implicit_run_steps_condition() -> None:
    self.runtime.start(run_condition=3)
    data = self.get_data(sink)
    self.assertEqual(data, [0, 1, 2, 3, 4, 5, 6, 7, 8, 0])
```

mathisrichter: And unless you expect users to use the `runtime.start()` method, I would test on the Process-level API
```python
source = SourceBuffer(data=np.array(range(1000)).reshape((1, -1)))
source.run(run_cfg=Loihi2SimCfg(), condition=5)
```

src/lava/magma/core/process/process.py
```diff
@@ -304,7 +304,7 @@ def is_sub_proc_of(self, proc: AbstractProcess):
             return False
 
     def run(self,
-            condition: AbstractRunCondition,
+            condition: ty.Optional[ty.Union[AbstractRunCondition, int]],
```
mathisrichter, line 307: Can we fix this inconsistency while we are at it? Why is one parameter called `condition` and the other `run_cfg`? Any of these combinations would make more sense to me:
- `config` and `condition` (prefer this)
- `run_config` and `run_condition` (okay)
- `run_cfg` and `run_cndtn` (ugh)

I also always get confused about which is which and look it up. Maybe there is a better naming entirely.

src/lava/magma/core/process/process.py
```diff
@@ -333,8 +333,10 @@ def run(self,
 
         Parameters
         ----------
-        condition : AbstractRunCondition
-            Specifies for how long to run the Process.
+        condition : Optional[Union[AbstractRunCondition, int]], default=None
```
mathisrichter, line 336: Just had this thought: When we have a Process with a `PyProcModel` that implements the `AsyncProtocol`, are we treating an iteration of the `run()` method as a "timestep"? I'm generally confused about these global time steps of the `RunSteps` condition when dealing with a network of Processes that run asynchronously.

src/lava/magma/runtime/runtime.py
```diff
@@ -327,16 +327,27 @@ def _get_resp_for_run(self):
...
+              ty.Optional[ty.Union[AbstractRunCondition, int]] = None):
         """
-        Given a run condition, starts the runtime
-
-        :param run_condition: AbstractRunCondition
-        :return: None
+        Start the runtime and run as long as the run_condition dictates.
+        Parameters
+        ----------
+        run_condition : Optional[Union[AbstractRunCondition, int]],
```
mathisrichter, line 336: Here it is called `run_condition` rather than `condition`.

src/lava/magma/core/process/process.py
mathisrichter, line 336: I am a little concerned that it may not be obvious what `condition=10` means when they see this in a tutorial without context. Without knowing what a `RunCondition` is, they may think 10 is some magic number.

This is my main concern with this whole feature. Maybe it can be clarified by renaming the parameter?

mathisrichter (changes requested):
Generally very nice - I am just a little concerned about the parameter naming. I think that needs to be revisited now that it is possible to go without a specific `RunConfig` object.

PhilippPlank:
I am not sure if I like this change. Calling just `<some_proc>.run()` does not tell what is going to happen. Also `<some_proc>.run(10)` or `<some_proc>.run(condition=10)`. 

Maybe we can have parameters like:
`<some_proc>.run(steps=10)`
or
`<some_proc>.run(continuous=True)` (although False would not really make sense, but I cannot think of a good way to start a continuous run without importing something)

src/lava/magma/core/process/process.py
tim-shea: I agree completely and have considered making this change several times, unfortunately this will be breaking for user code, so we'd need to do a deprecation thing and I put it off before.

I'll look into whether it's possible to deprecate just an argument (run_cfg) and add config.

src/lava/magma/runtime/runtime.py
tim-shea: Yep.

tim-shea:
> I am not sure if I like this change. Calling just `<some_proc>.run()` does not tell what is going to happen. Also `<some_proc>.run(10)` or `<some_proc>.run(condition=10)`.

I agree that `proc.run(10)` is a bit weird and it would be better to add a param called steps. Agree to disagree that `proc.run()` is ambiguous. It does exactly what it says on the tin: runs.

> Maybe we can have parameters like: `<some_proc>.run(steps=10)` or `<some_proc>.run(continuous=True)` (although False would not really make sense, but I cannot think of a good way to start a continuous run without importing something)
> 
> Also blocking and non-blocking mode need to be specified for steps, which is currently part of the condition.

Adding blocking also as an argument seems like it would resolve these.

- `proc.run()` would be non-blocking, continuous.
- `proc.run(blocking=False)` same as previous
- `proc.run(blocking=True)` would be blocking, continuous. Perhaps should log a warning or info along the lines of "Running a process in blocking, continuous mode. Note that one or more connected processes must request a stop or pause or this will run forever."
- `proc.run(steps=N)` would be blocking, num-steps
- `proc.run(steps=N, blocking=True)` same as previous
- `proc.run(steps=N, blocking=False)` would be non-blocking, num-steps

Someone long ago made the decisions that RunSteps should default to blocking while RunContinuous should default to non-blocking, and while I find that not exactly obvious it has a bit of logic to it and I'm hesitant to change reasonable behavior without a good reason.

src/lava/magma/core/process/process.py
joyeshmishra: I would also not prefer to overload. Add another API like run_for_timestep(...) and pipe that through this generic run API. But let's not overload the existing API.

src/lava/magma/runtime/runtime.py
joyeshmishra: start should not even do the overload. This would be confusing. a the earlier it meant it will only start and not run the experiment. Now, passing the int will make the purpose of the start API confusing.

src/lava/magma/core/process/process.py
mathisrichter: Do we really need to deprecate something at this stage? I feel we could start with that when reaching Lava version 1.0...

tim-shea:
This isn't necessary. It was just a nice to have.

Issue #770, opened 2023-08-06
Make it easier to use Buffers for input and output


tim-shea:
### User story
As a user, I want to be able to store or write multiple inputs, outputs, or vars easily, without creating many different RingBuffer processes.

### Conditions of satisfaction
- Allow a buffer to dynamically create ports and vars when it is connected.
- Ensure that multiple buffers with different connectivity do not interfere with each other (e.g. due to the use of class members).
- Provide flexible, user-configurable options for what to do when a buffer overflows:
  - Raise an error, since we may assume the user intended to match the length of a buffer to the number of timesteps
  - Wrap around, as with the current RingBuffers
  - Reallocate, e.g. by doubling the available memory in the buffer (as in C++ standard library containers)
- Allow the user to create and connect a buffer without any duplication of shape parameters from the connected ports

PR #771, opened 2023-08-06
Create a new Buffer class that can flexibly combine the functions of current RingBuffer, Read, and Write.

tim-shea:
Issue Number: #770 

Objective of pull request: Create a new class `Buffer` that will dynamically create ports and vars as needed to allow the user to connect arbitrary inputs, outputs, and vars to pre-allocated numpy memory buffers. This will significantly simplify the the I/O package and make it easier to stimulate or record from networks.

## Pull request checklist
-   [X] [Issue](https://github.com/lava-nc/lava/issues) created that explains the change and why it's needed
-   [X] Tests are part of the PR (for bug fixes / features)
-   [X] [Docs](https://github.com/lava-nc/docs) reviewed and added / updated if needed (for bug fixes / features)
-   [X] PR conforms to [Coding Conventions](https://lava-nc.org/developer_guide.html#coding-conventions)
-   [X] [PR applys BSD 3-clause or LGPL2.1+ Licenses](https://lava-nc.org/developer_guide.html#add-a-license) to all code files
-   [X] Lint (`flakeheaven lint src/lava tests/`) and (`bandit -r src/lava/.`) pass locally
-   [X] Build tests (`pytest`) passes locally

## Pull request type
-   [X] Feature

-   [X] No

### DIFF
--- a/src/lava/proc/io/buffer.py
+++ b/src/lava/proc/io/buffer.py
@@ -1,0 +1,182 @@
+# Copyright (C) 2021-22 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+# See: https://spdx.org/licenses/
+
+import numpy as np
+
+from typing import Optional, Union, Type
+
+from lava.magma.core.process.variable import Var
+from lava.magma.core.process.process import AbstractProcess
+from lava.magma.core.process.ports.ports import InPort, OutPort, RefPort
+
+from lava.magma.core.resources import CPU
+from lava.magma.core.decorator import implements, requires, tag
+from lava.magma.core.model.py.model import PyLoihiProcessModel
+from lava.magma.core.sync.protocols.loihi_protocol import LoihiProtocol
+from lava.magma.core.model.py.type import LavaPyType
+from lava.magma.core.model.py.ports import PyInPort, PyOutPort, PyRefPort
+
+
+class Buffer(AbstractProcess):
+    """Buffer receives data from OutPorts or VarPorts in each
+    timestep.
+
+    To add a connection, call `connect` with either an InPort
+    or a Var from another process.
+
+    To read the contents of the buffer, call `get` on the Var
+    returned by a call to `connect` while the process model
+    is still active (i.e. before calling `process.stop`).
+
+    TODO: Implement 'wrap_around', 'reallocate', 'read_write_file'
+    TODO: Support OutPorts also
+
+    Parameters
+    ----------
+    length: int, default = 100
+        The number of timesteps that can be recorded in the buffer.
+    overflow: str, default = 'raise_exception'
+        The desired behavior when the buffer overflows. Options are
+        'raise_exception', 'wrap_around', and 'reallocate'.
+    """
+    def __init__(self,
+                 *,
+                 length: int = 100,
+                 overflow: str = 'raise_error') -> None:
+        super().__init__(length=length, overflow=overflow,
+            map_out=[], map_in=[], map_ref=[])
+        self.length = length
+        self.overflow = overflow
+        self.map_out = []
+        self.map_in = []
+        self.map_ref = []
+        self.index = 1000
+
+    def connect(self, other: Union[InPort, OutPort, Var],
+                init: Optional[np.ndarray] = None) -> Var:
+        """Connect the buffer to an OutPort or Var from another process.
+        Calling this method will instantiate a new InPort or RefPort as
+        needed in the buffer and a corresponding Var of the appropriate
+        shape and length.
+        Parameters
+        ----------
+        other: Union[OutPort, Var]
+            The other port or var to connect to and store in the buffer.
+        init: Optional[ndarray]
+            The initial value of the buffer Var. This will determine the
+            values sent from an InPort buffer and the default values for
+            an OutPort or RefPort buffer.
+        Returns
+        -------
+        The Var which will store the buffered data.
+        """
+        index = self.index
+        var_shape = other.shape + (self.length,)
+        if init is None:
+            init = np.zeros(var_shape)
+        var = Var(shape=var_shape, init=init)
+        var.name = f'Var{index}'
+        setattr(self, var.name, var)
+
+        if isinstance(other, InPort):
+            port = OutPort(shape=other.shape)
+            port.name = f'Out{index}'
+            other.connect_from(port)
+            self.map_out.append((var.name, port.name))
+            self.proc_params.overwrite('map_out', self.map_out)
+        elif isinstance(other, OutPort):
+            port = InPort(shape=other.shape)
+            port.name = f'In{index}'
+            other.connect(port)
+            self.map_in.append((var.name, port.name))
+            self.proc_params.overwrite('map_in', self.map_in)
+        elif isinstance(other, Var):
+            port = RefPort(shape=other.shape)
+            port.name = f'Ref{index}'
+            port.connect_var(other)
+            self.map_ref.append((var.name, port.name))
+            self.proc_params.overwrite('map_ref', self.map_ref)
+        else:
+            raise ValueError(f'Other {other} is not an InPort, OutPort, '
+                             'or Var.')
+        setattr(self, port.name, port)
+        self._post_init()
+        self.index += 1
+        return var
+
+
+class MetaPyBuffer(type(PyLoihiProcessModel)):
+    """This metaclass allows dynamic port and var generation."""
+    def __getattr__(cls, name):
+        if 'In' in name:
+            return LavaPyType(PyInPort.VEC_DENSE, float)
+        elif 'Out' in name:
+            return LavaPyType(PyOutPort.VEC_DENSE, float)
+        elif 'Ref' in name:
+            return LavaPyType(PyRefPort.VEC_DENSE, float)
+        elif 'Var' in name:
+            return LavaPyType(np.ndarray, float)
+        else:
+            raise AttributeError(name=name, obj=cls)
+
+
+@implements(proc=Buffer, protocol=LoihiProtocol)
+@requires(CPU)
+class PyBuffer(PyLoihiProcessModel, metaclass=MetaPyBuffer):
+    """Python CPU model for Buffer. Uses dense floating point numpy
+    arrays for buffer storage and operations."""
+    def __init__(self, proc_params):
+        super().__init__(proc_params)
+        self.length = proc_params['length']
+        self.overflow = proc_params['overflow']
+        self.map_in = proc_params['map_in']
+        self.map_out = proc_params['map_out']
+        self.map_ref = proc_params['map_ref']
+
+        for var, port in self.map_in:
+            setattr(self, var, LavaPyType(np.ndarray, float))
+            setattr(self, port, LavaPyType(PyInPort.VEC_DENSE, float))
+
+        for var, port in self.map_out:
+            setattr(self, var, LavaPyType(np.ndarray, float))
+            setattr(self, port, LavaPyType(PyOutPort.VEC_DENSE, float))
+
+        for var, port in self.map_ref:
+            setattr(self, var, LavaPyType(np.ndarray, float))
+            setattr(self, port, LavaPyType(PyRefPort.VEC_DENSE, float))
+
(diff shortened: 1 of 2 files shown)


src/lava/proc/io/buffer.py
```diff
@@ -0,0 +1,182 @@
...
+        super().__init__(length=length, overflow=overflow,
+            map_out=[], map_in=[], map_ref=[])
+        self.length = length
+        self.overflow = overflow
+        self.map_out = []
+        self.map_in = []
+        self.map_ref = []
+        self.index = 1000
+
+    def connect(self, other: Union[InPort, OutPort, Var],
```
mathisrichter, line 56: It may be helpful to split this into a `connect()` and a `connect_from()` method, as we are doing it for ports. As it is now, I'm having trouble understanding the direction in which data flow will flow.

tests/lava/proc/io/test_buffer.py
```diff
@@ -0,0 +1,118 @@
...
+        self.assertSequenceEqual(data, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+
+    def test_buffer_with_2vars(self):
+        """Test to ensure that a Buffer can connect to multiple Vars."""
+        num_steps = 10
+        injector = Injector(shape=(1,))
+        lif = LIF(shape=(1,), du=1)
+        buffer = Buffer(length=10)
+        injector.out_port.connect(lif.a_in)
+        u = buffer.connect(lif.u)
```
mathisrichter, line 68: So this creates a `Var u` but does not place it inside the `buffer` object? I was somehow not expecting that.

mathisrichter (changes requested):
I may be misunderstanding this. At least for output from a network, this seems to be overlapping with the `Monitor` Process. Would it make sense to add any output-related functionality to the `Monitor`?

I am somewhat confused by the API. It would expose Lava `Vars` to the application scope, independent of the Process that it was generated from. This may break up the concept of a Process too much. Maybe there is a way to add the Vars to the `Buffer` Process? I'm guessing you did not do this on purpose - let's discuss why. :)

tim-shea:
> I may be misunderstanding this. At least for output from a network, this seems to be overlapping with the `Monitor` Process. Would it make sense to add any output-related functionality to the `Monitor`?

This is basically a refactor of the code from the current RingBuffer, Read, and Write processes. Read and Write definitely do overlap with Monitor, and it would be interesting to better understand how to eliminate that duplication. I started a refactor on Monitor a few months ago to turn it into a process with proper ports, but didn't make it far enough to commit anything. I'd be very interested to have a discussion to better understand Monitor architecture and try to reconcile these different I/O bits.

> I am somewhat confused by the API. It would expose Lava `Vars` to the application scope, independent of the Process that it was generated from. This may break up the concept of a Process too much. Maybe there is a way to add the Vars to the `Buffer` Process? I'm guessing you did not do this on purpose - let's discuss why. :)

I don't know what you mean by 'application scope'. The goal here is not to alter the extent to which Vars are exposed, only to simplify the usage of the existing Read and Write processes which act like input and output buffers for Vars. Are you familiar with Read and the PyRead models in source.RingBuffer? This should replace those.

src/lava/proc/io/buffer.py
tim-shea: Yep, that would make sense. I'll make that change.

I don't love the names `connect` and `connect_from` but they're not awful and this would better differentiate the behavior of connecting to an InPort or Var (which will send the values from init) vs connecting from an OutPort or Var (in which case init will generally be overwritten).

tests/lava/proc/io/test_buffer.py
tim-shea: I can see how this is slightly confusing, because the use of dynamic port and var creation doesn't allow you to immediately see what's happening in PyBuffer, but the port and vars *are* added to the Buffer process and the corresponding PyPort and PyVars are added to the PyBuffer process model.

They can be accessed in the buffer.in_ports, buffer.out_ports, buffer.ref_ports, and buffer.vars collections, or they can be accessed directly with the use of a name as an attribute of the buffer. Unfortunately, I hit a small snag where I couldn't quite find a way to achieve user-specified names for those objects while also making it easy to intercept calls to the metaclass.getattr (necessary to convince the compiler to allow the process model to be built, but that's sort of a hack that could later be fixed in the compiler). Thus, the name of the var that will be used as a buffer is currently `VarXXXX` with the x's pulled from a incrementing sequence starting at 1000.

So, you could do:
```
buffer = Buffer()
u = buffer.connect(some_other_proc.some_port)
assert(u == buffer.Var1000)
```

I'll think a bit more on whether there's a way to let the user specify the name of the var, as I do find that more compelling than `Var1000`.

tim-shea:
> I am somewhat confused by the API. It would expose Lava `Vars` to the application scope, independent of the Process that it was generated from. This may break up the concept of a Process too much. Maybe there is a way to add the Vars to the `Buffer` Process? I'm guessing you did not do this on purpose - let's discuss why. :)

Ahh I think I figured out what you mean by exposing vars while responding to your other question. In the sense that my application code could end up with a Var called u, yes I see how that looks like it breaks down the encapsulation. The thing is, that reference to u is not the same u from the LIF process, it's a buffer into which u (the one from the LIF) will get written on each timestep. We return that var just to get around a bit naming ugliness (explained above) by giving the user a handy reference to their buffered data. Since the buffered version of u is *always* either input data specified by the user or output data which will be manipulated by the user, I don't see any issue with encapsulation there, although I suppose we could in theory not return the Var in the input case, since the initialization data goes in the method call (but then, what if I want to rerun with different data?) and return a Future (in the output case) to ensure the semantics of what you should do with "u" are clear.

src/lava/proc/plot/streaming.py
gkarray, line 1: Would it make sense to add a small script/tutorial showing how to use the classes defined in this file ?
It doesn't have to be in this PR I guess, since it's not directly related to the Buffer concept you're introducing here.

tests/lava/proc/io/test_buffer.py
```diff
@@ -0,0 +1,129 @@
...
+        self.assertTrue(hasattr(buffer, 'v'))
+        self.assertEqual(buffer.v.shape, buffer_shape)
+        self.assertTrue(buffer.vInPort0 in buffer.in_ports)
+
+    def test_run_buffer(self):
+        num_steps = 10
+        injector = Injector(shape=(1,))
+        buffer = Buffer(length=10)
+        buffer.connect_from('v0', injector.out_port)
+        buffer.create_runtime(run_cfg=Loihi2SimCfg())
```
gkarray, line 63: Is it necessary to explicitly call `buffer.create_runtime` before running ?
If so, please add a mention of this in the docstring of the Buffer class otherwise it might not be very straightforward for users to find out about it.

tests/lava/proc/io/test_buffer.py
```diff
@@ -0,0 +1,129 @@
...
+from lava.magma.core.process.variable import Var
+from lava.magma.core.run_configs import Loihi2SimCfg
+from lava.magma.core.run_conditions import RunSteps
+from lava.proc.io.injector import Injector
+from lava.proc.io.extractor import Extractor
+from lava.proc.lif.process import LIF
+from lava.proc.io.buffer import Buffer
+
+
+class TestBuffer(unittest.TestCase):
```
gkarray, line 19: For consistency with how we test other Processes and ProcessModels in Lava, we might want to split tests into TestProcess and TestProcessModel. Honestly, I don't like that divide myself, and I prefer it the way you did it here...
But I think it is crucial for code quality to be consistent in how we test these things.
So if we decide to go this route and drop the divide between TestProcess and TestProcessModel in the future, we should at least write an issue about the need to refactor other Process tests.

tests/lava/proc/io/test_buffer.py
```diff
@@ -0,0 +1,129 @@
...
+        buffer = Buffer(length=10)
+        var_shape = (3,)
+        buffer_shape = (3, 10)
+        test_out_port = OutPort(shape=var_shape)
+        buffer.connect_from('v', test_out_port)
+        self.assertTrue(hasattr(buffer, 'v'))
+        self.assertEqual(buffer.v.shape, buffer_shape)
+        self.assertTrue(buffer.vInPort0 in buffer.in_ports)
+
+    def test_run_buffer(self):
```
gkarray, line 58: For consistency with other test names, this should be something like `test_run_single_input_buffer`.

tests/lava/proc/io/test_buffer.py
```diff
@@ -0,0 +1,129 @@
...
+        buffer.create_runtime(run_cfg=Loihi2SimCfg())
+        buffer.run(condition=RunSteps(num_steps, blocking=False))
+        for t in range(num_steps):
+            injector.send(np.full((1,), t))
+        buffer.wait()
+        data = buffer.v0.get().flatten().astype(int).tolist()
+        buffer.stop()
+        self.assertSequenceEqual(data, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+
+    def test_buffer_with_2vars(self):
```
gkarray, line 72: For consistency with other test names, this should be something like `test_run_single_input_buffer_2_vars`.

tests/lava/proc/io/test_buffer.py
```diff
@@ -0,0 +1,129 @@
...
+        for t in range(num_steps):
+            injector.send(np.full((1,), t))
+        buffer.wait()
+        udata = buffer.u.get().flatten().astype(int).tolist()
+        vdata = buffer.v.get().flatten().astype(int).tolist()
+        buffer.stop()
+        self.assertSequenceEqual(udata, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+        self.assertSequenceEqual(vdata, [0, 1, 3, 6, 10, 0, 6, 0, 8, 0])
+
+    def test_multiple_buffers(self):
```
gkarray, line 92: For consistency with other test names, this should be something like `test_run_multiple_input_buffer`.

tests/lava/proc/io/test_buffer.py
```diff
@@ -0,0 +1,129 @@
...
+        for t in range(num_steps):
+            injector.send(np.full((1,), t))
+        ubuffer.wait()
+        udata = ubuffer.u.get().flatten().astype(int).tolist()
+        vdata = vbuffer.v.get().flatten().astype(int).tolist()
+        ubuffer.stop()
+        self.assertSequenceEqual(udata, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+        self.assertSequenceEqual(vdata, [0, 1, 3, 6, 10, 0, 6, 0, 8, 0])
+
+    def test_output_buffer(self):
```
gkarray, line 114: For consistency with other test names, this should be something like `test_run_single_output_buffer`.
Oh, and don't forget the docstring for this test.

gkarray (changes requested):
Very great addition to the I/O package, thanks @tim-shea !
Nothing much to say about the implementation itself, except that I like this trick of dynamic Port and Var creation :).
Otherwise, I added a few minor comments about syntax.

tests/lava/proc/io/test_extractor.py
```diff
@@ -392,7 +399,11 @@ def _test_receive_not_empty_policy(receive_not_empty: utils.ReceiveNotEmpty,
...
 
         extractor.run(condition=run_condition, run_cfg=run_cfg)
+        if verbose:
+            print(f'{extractor.runtime._is_running=}')
         recv_data = [extractor.receive(), extractor.receive()]
+        if verbose:
+            print(f'{recv_data=}')
         extractor.stop()
 
         return np.array(recv_data)
```
mgkwill, line 409: `test_receive_data_receive_not_empty_fifo` and `test_receive_data_receive_not_empty_accumulate` appear to fail because `self.runtime._is_running` is not true at time of `extractor.receive()` and thus the check @tim-shea put in `can_receive()` fails, thus the correct `qsize` is not used when calling internal `_recieve_when*`.

``` 
def can_receive(self) -> int:
        if self.runtime is not None and \
            self.runtime._is_running:
            return self._pm_to_p_dst_port._queue.qsize()
        else:
            return 0
```

PR #839, opened 2024-02-29
Spiker with 32bit

phstratmann:
Adds a spiker with 32bit messages
Also, this spiker actually has a correct definition of period. It sends its first spike after 'period' number of time steps.


GaboFGuerra (approved):
Thanks Philipp! looks good :)

AlessandroPierro (approved):
Looks good, thank you!

ashishrao7 (approved):
Works as it should. Tested behavior

